PR #35, opened 2025-08-14
refactor: Make use of Tools using the OpenAI spec properly

edenreich:
## Summary

The CLI was using a weird Text based recognition for tool calls. This fixes it by using the Tool-Use API.

Also adding improved visual effects - expandable tool call results using ctrl+r similar to what we used to from Claude Code.


edenreich:
@claude Please address the Memory Management issue. The rest of the things are fine. Maybe to address it we can only keep track of the current active tool? I only able to expand the current tool executed (last tool executed) so maybe we don't need to track all of the tool results but rather just the last one?

PR #83, opened 2025-08-19
feat: Implement centralized sandbox configuration

### LINKED ISSUE #82: [TASK] Reconcile Sandbox environment
## Summary

Currently the .infer/config.yaml contains tools specific sandboxing. This is not ideal and sometimes even repetitive. Ideally there should be a command to setup the context of the project (the only context where the LLM maybe executes tools on) - default to current dir (when running init it should add that to the config).

If the LLM wants to execute a tool they should only be able to execute it in that configured directory. Let's allow a list of directories.

Let's not care about backward compatibility - delete code that is not needed.

### Acceptance Criteria

- [ ] There is a sandbox configurations to ensure the LLM can only executes tools in this environment
- [ ] Other path related configurations are removed from the tool-level config
- [ ] It's tested
- [ ] It's documented
- [ ] Old documentation is removed

edenreich:
Resolves #82

This PR implements centralized sandbox configuration to replace the current tool-specific sandboxing approach.

## Changes
- Add `SandboxConfig` with `Directories` field to replace tool-specific path configs
- Implement `ValidatePathInSandbox()` method for centralized validation
- Update all tools to use sandbox validation
- Remove `ExcludePaths`, `RestrictToWorkDir`, and `ProtectedPaths` configurations
- Add CLI commands: `infer config tools sandbox [list|add|remove]`
- Replace exclude-path commands with sandbox commands
- Set default sandbox to current directory in DefaultConfig
- Update tests and documentation

## Breaking Changes
As requested, this removes tool-level path configurations without maintaining backward compatibility.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -184,33 +184,33 @@ var configToolsSafetyUnsetCmd = &cobra.Command{
 	RunE:  unsetToolApproval,
 }
 
-var configToolsExcludePathCmd = &cobra.Command{
-	Use:   "exclude-path",
-	Short: "Manage excluded paths",
-	Long:  `Manage paths that are excluded from tool access for security purposes.`,
+var configToolsSandboxCmd = &cobra.Command{
+	Use:   "sandbox",
+	Short: "Manage sandbox directories",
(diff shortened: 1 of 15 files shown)


edenreich:
@claude Tests are still failing - please fix them.

Issue #110, opened 2025-08-23
[FEATURE] Improve System Prompt to Respect `AGENTS.md` by Default


edenreich:
## Summary

OpenAI recently introduced [`[AGENTS.md](https://github.com/openai/agents.md)`](https://github.com/openai/agents.md), which is quickly becoming the de-facto standard for providing markdown-based instructions to LLMs.

To align with this standard, the system prompt should:

1. Prefer `AGENTS.md` when available in the repository.
2. Fall back to `README.md` if no `AGENTS.md` file exists.

This ensures a more consistent and efficient coding experience, since LLMs can parse structured, instruction-focused content rather than general project documentation.

## Background

Previously, each LLM vendor introduced their own convention for project-specific instructions:

* `GEMINI.md` for Gemini
* `QWEN.md` for Qwen
* `CLAUDE.md` for Claude
* and similar vendor-specific files

This fragmented ecosystem created friction for developers and tools trying to support multiple providers. `AGENTS.md` consolidates these efforts into a single, vendor-agnostic format—making it easier to maintain and ensuring better interoperability.

## Motivation

Currently, each LLM vendor still interprets prompts and documentation differently. Adopting `AGENTS.md` as the default source of truth will:

* Encourage consistency across tools.
* Reduce confusion caused by vendor-specific quirks.
* Future-proof the workflow as more providers adopt the standard.

In the meantime, a vendor-aware fallback (dynamic system prompt adaptation) may be needed until adoption is universal.

## Acceptance Criteria

* [ ] If `AGENTS.md` is present, system prompt automatically references it.
* [ ] If `AGENTS.md` is missing, system prompt falls back to `README.md`.
* [ ] Optional: Vendor-specific handling for providers not yet supporting the standard.

edenreich:
@claude Can you implement it?

PR #113, opened 2025-08-24
feat(config): Prefer AGENTS.md over README.md in system prompt

edenreich:
Updates the default system prompt to align with OpenAI's AGENTS.md standard:
- Changes instruction to prefer AGENTS.md when available
- Maintains fallback to README.md for backwards compatibility
- Updates corresponding test expectations

Addresses issue #110 to adopt the vendor-agnostic AGENTS.md convention for project-specific LLM instructions.


edenreich:
I think I'll not pre-processes the AGENTS.md file because I want to keep the agent dynamic, this agent might run on multiple repositories during a single task and each repository will have their own AGENTS.md. Pre-processing the MD might be unnecessary, will let the agent decide when to read it. I might just need to specify in the System prompt what sections there might be inside of the file, so it can query those sections more effectively.

PR #163, opened 2025-09-21
feat: Add background task execution for A2A agents

### LINKED ISSUE #160: [FEATURE] Allow tasks to run completely in the background
## Summary

Currently when the CLI submits a task it immediately blocking and polling the A2A server.
 
Ideally when a task is submitted it should not block the A2A Client so the client (this CLI) can then perform other stuffs when necessary (Read, Write, Edit, MultiEdit locally etc, until the other task is completed and propagated back to the main context).

It's especially useful for scenarios like:

User ask about to create a workflow for n8n -> assistant submit a task to n8n agent to figure out how the workflow should be -> meanwhile the agent can for example create a folder where it would place -> the task arrived with the final artifact -> assistant place it in the directory it has created (things could be parallelized - the whole purpose of tasks)

This should be an option offered to the user working on the terminal by for example pressing ctrl + b (b for background).

The conversation should remain in tact and in the correct order, so for example once the task that was sent is back it should be added to the main context with the prior assistant message that has tool_calls[].id to ensure data integrity when submitting this back to the provider's APIs. 

Thing to consider - maybe instead of manually polling I could simply provide a tool for the LLM to decide when it wants to poll it with a minimum threshold (probably not cost effective, but worth considering, because it makes the agent fully autonomous).

This should support only the Polling strategy not the Push notifications - I'm not sure I want to implement the push notifications because it requires opening a door on the A2A Client (webhook).

### Acceptance Criteria

- [ ] Task could be sent off to the background - allowing the CLI to perform other tool calls meanwhile
- [ ] It's documented
- [ ] It's tested

WIP: I need to refine this ticket.

edenreich:
Implements background task execution for A2A agents allowing non-blocking operation and concurrent local tasks.

## Features
- Background task execution with `run_in_background` parameter
- ctrl+b key binding for background task management
- Status bar indicators showing active task count
- Automatic completion notifications
- Conversation history integration with proper threading
- Configurable polling intervals

## Architecture
- BackgroundTaskManager service for lifecycle management
- Event-driven UI integration
- Thread-safe operations with proper cancellation
- Clean architecture with dependency injection

Fixes #160


edenreich:
@claude CI Is failing - please fix.

edenreich:
@claude There are still lint errors in the CI - can you fix them too?

PR #176, opened 2025-09-24
fix(a2a): Resolve A2A tools not available when using INFER_A2A_ENABLED=true

### LINKED ISSUE #174: [BUG] A2A Tools are not available in chat
## Summary

Following this docker compose service and running it I can't see any tools available from A2A:

```yaml
  cli:
    image: ghcr.io/inference-gateway/cli:latest
    pull_policy: always
    environment:
      INFER_LOGGING_DEBUG: true
      INFER_GATEWAY_URL: http://inference-gateway:8080
      INFER_A2A_ENABLED: true
      INFER_TOOLS_ENABLED: false
      INFER_AGENT_MODEL: deepseek/deepseek-chat
      INFER_A2A_AGENTS: |
        http://agent:8080
    command:
      - chat
    networks:
      - a2a-network
    profiles:
      - manual
```

It only works when explicitly setting the tools specific environment variables like `INFER_A2A_TOOLS_SUBMIT_TASK_ENABLED=true` etc

Expected is that when enabling A2A by default all of those tools are available unless otherwise specified.

edenreich:
Fixes #174

The issue was that when `INFER_A2A_ENABLED=true` was set without a config file, Viper's nested structure handling was causing individual A2A tool configurations to lose their default values during unmarshaling.

## Root Cause

- Using `v.SetDefault("a2a", defaults.A2A)` set the entire nested structure as one default
- When `v.Set("a2a.enabled", enabled)` was called, it created a partial structure that overwrote defaults
- During unmarshaling, individual tool configs were missing

## Solution

- Set A2A defaults individually using `v.SetDefault()` for each nested field
- Remove manual `v.Set("a2a.enabled", enabled)` and let `AutomaticEnv()` handle `INFER_A2A_ENABLED`
- Ensures proper environment variable precedence without losing default values

## Expected Behavior

- `INFER_A2A_ENABLED=true` → All A2A tools available by default
- Individual tools can still be disabled with specific env vars
- Works with or without config files

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -57,7 +57,23 @@ func initConfig() {
 	v.SetDefault("client", defaults.Client)
 	v.SetDefault("tools", defaults.Tools)
 	v.SetDefault("agent", defaults.Agent)
-	v.SetDefault("a2a", defaults.A2A)
+	v.SetDefault("compact", defaults.Compact)
+	v.SetDefault("git", defaults.Git)
+	v.SetDefault("storage", defaults.Storage)
+	v.SetDefault("conversation", defaults.Conversation)
+	v.SetDefault("chat", defaults.Chat)
+
+	// Set A2A defaults individually to ensure proper env var override behavior
+	v.SetDefault("a2a.enabled", defaults.A2A.Enabled)
+	v.SetDefault("a2a.agents", defaults.A2A.Agents)
+	v.SetDefault("a2a.cache", defaults.A2A.Cache)
+	v.SetDefault("a2a.task", defaults.A2A.Task)
+	v.SetDefault("a2a.tools.query_agent.enabled", defaults.A2A.Tools.QueryAgent.Enabled)
+	v.SetDefault("a2a.tools.query_agent.require_approval", defaults.A2A.Tools.QueryAgent.RequireApproval)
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -57,7 +57,23 @@ func initConfig() {
...
 	v.SetDefault("tools", defaults.Tools)
 	v.SetDefault("agent", defaults.Agent)
-	v.SetDefault("a2a", defaults.A2A)
+	v.SetDefault("compact", defaults.Compact)
+	v.SetDefault("git", defaults.Git)
+	v.SetDefault("storage", defaults.Storage)
+	v.SetDefault("conversation", defaults.Conversation)
+	v.SetDefault("chat", defaults.Chat)
+
+	// Set A2A defaults individually to ensure proper env var override behavior
```
edenreich, line 66:
```suggestion
```

cmd/root.go
```diff
@@ -68,6 +83,7 @@ func initConfig() {
 	v.AutomaticEnv()
 	v.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 
+	// Handle special A2A environment variables that need manual processing
```
edenreich, line 86:
```suggestion
```

PR #218, opened 2025-11-21
feat(gateway): Add progress feedback for gateway initialization

### LINKED ISSUE #217: [FEATURE] Add better UX when entering the chat session or running in agent mode
## Summary

Currently, when docker is configured for the gateway, it sometimes take time to get into the session and the user won't know what happens behind the scene - it simply hangs because:
1. It download the gateway container image (even thought it's only ~15mb to download it still depends on networking and I should give a download feedback to the user)
2. It runs the container (could also take a few ms)

When disabling docker the CLI is running instantly with the binary gateway (but probably still it needs to download the gateway binary on the first run which can take a while - not as slow as docker but still slow I need to give feedback for the download). 

### Acceptance Criteria

- [ ] There is a feedback for the Gateway download if gateway.run is set to true
- [ ] There is no feedback for the download if the gateway.run is not set to true (because it's connected to remote - there is no download)
- [ ] The overall UX is improved
- [ ] The existing attributes are documented (if applicable)

edenreich:
## Summary

Adds clear visual progress indicators during gateway initialization to improve UX and prevent confusing hangs.

## Changes

- Add visual progress indicators during gateway startup
- Show download progress for Docker image pulls
- Show download progress for binary gateway downloads
- Display status messages for container/binary startup
- Add ready confirmation with gateway URL
- Only show feedback when gateway.run is enabled

## Testing

- All tests pass
- Linting: 0 issues
- Build successful

Resolves #217

---

### DIFF
--- a/internal/services/gateway_manager.go
+++ b/internal/services/gateway_manager.go
@@ -182,12 +182,10 @@ func (gm *GatewayManager) IsRunning() bool {
 func (gm *GatewayManager) pullImage(ctx context.Context) error {
 	logger.Debug("Pulling gateway image", "image", gm.config.Gateway.OCI)
 
-	// Print progress message to user
 	fmt.Printf("• Pulling gateway image: %s\n", gm.config.Gateway.OCI)
 
 	cmd := exec.CommandContext(ctx, "docker", "pull", gm.config.Gateway.OCI)
 
-	// Capture output but also show progress
 	output, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("docker pull failed: %w, output: %s", err, string(output))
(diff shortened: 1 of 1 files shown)


internal/services/gateway_manager.go
```diff
@@ -165,14 +178,22 @@ func (gm *GatewayManager) IsRunning() bool {
...
+// pullImage pulls the OCI image with progress feedback
 func (gm *GatewayManager) pullImage(ctx context.Context) error {
 	logger.Debug("Pulling gateway image", "image", gm.config.Gateway.OCI)
+
+	// Print progress message to user
+	fmt.Printf("• Pulling gateway image: %s\n", gm.config.Gateway.OCI)
+
 	cmd := exec.CommandContext(ctx, "docker", "pull", gm.config.Gateway.OCI)
+
+	// Capture output but also show progress
```
edenreich:
```suggestion
```

PR #233, opened 2025-11-22
feat: Add image attachment support with vision model detection

edenreich:
This PR adds image attachment support with vision model detection:

- Added image service integration to handle image files
- Implemented image attachment functionality in chat input
- Added vision model detection to warn users when using non-vision models with images
- Updated file service to handle image files with appropriate size limits
- Modified chat message processor to handle image references
- Added support for image files in the Read tool with proper error handling

The changes allow users to attach images to their chat messages and provide warnings if the current model doesn't support vision capabilities.

### DIFF
--- a/internal/handlers/chat_message_processor.go
+++ b/internal/handlers/chat_message_processor.go
@@ -55,10 +55,8 @@ func (p *ChatMessageProcessor) handleUserInput(
 		}
 	}
 
-	// Merge images from @ references with any existing clipboard images
 	allImages := append(msg.Images, result.images...)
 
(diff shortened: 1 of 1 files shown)


internal/handlers/chat_message_processor.go
```diff
@@ -49,7 +55,30 @@ func (p *ChatMessageProcessor) handleUserInput(
 		}
 	}
 
-	return p.processChatMessage(expandedContent, msg.Images)
+	// Merge images from @ references with any existing clipboard images
```
edenreich:
```suggestion
```

PR #265, opened 2025-11-26
refactor(config): Replace agent.optimization with compact config

### LINKED ISSUE #264: [TASK] Refactor /compact reminder
## Summary

Since Estimated Context Window length is available, the configuration should simply use percentage. How far in the conversation should the compact get triggered in percentage.

The Task is to refactor the config and only provide `compact.auto_at: 10`, value from 20 to 100.

Also when pressing /compact it should execute immediately and not get queued.

### Acceptance Criteria

- [ ] The shortcut /compact executes immediately when pressed enter - not queued
- [ ] The config contains compact.auto_at which will take a percentage value when the conversation should get automatically compacted
- [ ] It's tested
- [ ] It's documented
- [ ] The config contains compact.enabled flag and is disabled by default
- [ ] The config replaces agent.optimization section - this section is deleted (no need backward compatibility, breaking changes are fine, it's not stable yet).

edenreich:
## Summary

Refactors the conversation compaction system to use percentage-based triggering instead of message count.

- Replaces `agent.optimization` config section with `compact` config
- Adds `compact.enabled` flag (default: false) and `compact.auto_at` percentage (default: 80)
- `/compact` shortcut now executes immediately instead of being queued
- Automatic compaction triggers based on percentage of context window used
- Removes old `config optimization` CLI commands

## Breaking Changes

The `agent.optimization` config section has been removed. Users must update their config to use `compact.enabled` and `compact.auto_at` instead.

## Test Plan

- [x] All existing tests pass
- [x] Code builds successfully
- [x] Documentation updated

Closes #264

### DIFF
--- a/internal/ui/keybinding/actions.go
+++ b/internal/ui/keybinding/actions.go
@@ -520,13 +520,21 @@ func handleCancel(app KeyHandlerContext, keyMsg tea.KeyMsg) tea.Cmd {
 
 	stateManager := app.GetStateManager()
 
-	if stateManager.GetCurrentView() == domain.ViewStatePlanApproval {
-		planApprovalState := stateManager.GetPlanApprovalUIState()
-		if planApprovalState != nil && planApprovalState.ResponseChan != nil {
-			return func() tea.Msg {
-				return domain.PlanApprovalResponseEvent{
-					Action: domain.PlanApprovalReject,
-				}
+	planApprovalState := stateManager.GetPlanApprovalUIState()
+	if planApprovalState != nil {
+		return func() tea.Msg {
+			return domain.PlanApprovalResponseEvent{
+				Action: domain.PlanApprovalReject,
+			}
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


internal/services/conversation_optimizer.go
```diff
@@ -7,45 +7,51 @@ import (
...
+	Config     *config.Config
+	Tokenizer  *TokenizerService
 }
 
 // NewConversationOptimizer creates a new conversation optimizer with configuration
 func NewConversationOptimizer(config OptimizerConfig) *ConversationOptimizer {
-	if config.MinMessages <= 0 {
-		config.MinMessages = 10
+	if config.AutoAt < 20 || config.AutoAt > 100 {
+		config.AutoAt = 80
```
edenreich: In this if block there should be a warning log indicating that AutoAt can only be between the range and that it is defaulting to 80 because the value that was set is incorrect.

edenreich:
@claude Please resolve the comments.

PR #342, opened 2025-12-16
feat: Add conversation versioning

### LINKED ISSUE #341: [FEATURE] Implement conversation versioning to enable going back in time
## Summary

Currently if the conversation / session is getting too many tokens in, you could end up in a situation where you can no longer compact the conversation but the context is out of bound. To solve this issue, what the user might want to do is press double esc and go back to a previous time of the conversation, allowing the user to "erase" the last overwhelming tokens that were pushed into the context, allowing the user to continue with the previously broken conversation.

### Acceptance Criteria

- [ ] There is a new key binding - defaulting to double esc
- [ ] When double esc is pressed the user is presented with truncated user's messages followed by an ordered list
- [ ] The dropdown shown to the user in the view allows the user to select a previous message of the user
- [ ] Selecting a point previous prompt simply deletes all the messages after it in the conversation - allowing the user to resume it
- [ ] It's documented
- [ ] It's tested

### Out of scope

Restoring code changes, this could be not very efficient in terms of storage, because I would have to take snapshots of every code change, native git is already fine for those use cases.

edenreich:
Adds conversation versioning feature that allows users to navigate back to previous conversation points. Users can press ESC twice to open a message history selector, choose a restore point, and delete all messages after that point.

## Key changes:

1. **Double ESC keybinding** - Triggers message history view (esc,esc)
2. **UI component** - Message history selector for choosing restore points from user message history
3. **Permanent deletion** - Deletes all messages after selected restore point
4. **Comprehensive documentation** - Feature documentation with examples and usage guide
5. **Unit tests** - Tests for ESC detection and message deletion functionality
6. **State management** - New view state and handlers for conversation versioning

## Feature details:

- **Trigger**: Double ESC (within 300ms) from chat view
- **Navigation**: Arrow keys to browse, `/` to search, Enter to select
- **Restore**: Deletes all messages after selected point, auto-saves changes
- **Supported modes**: Standard, Plan, and Auto-Accept agent modes
- **Edge cases**: Handles empty conversations, single messages, active sessions

## Technical implementation:

- Timer-based double ESC detection (300ms threshold)
- Thread-safe state management with mutex locks
- Repository-level message deletion with array truncation
- Dedicated view state (`ViewStateMessageHistory`)
- Auto-save integration for persistent storage

## Documentation:

- Complete usage guide with examples
- Keybinding configuration instructions
- FAQ section addressing common questions
- Technical details for developers

Closes #341


edenreich:
### TODOs

I haven't decided yet what shall I do with the calculated tokens when the user goes back in time - probably the estimated cost should remain and the token usage in the context should get re-calculated, but the pricing estimation is based on the token usage, so I'm not entirely sure yet what would be the best approach here.

PR #537, opened 2026-05-24
feat(ui): Show elapsed time and model in background A2A indicator

### LINKED ISSUE #536: [FEATURE] Add elapsed time and model name to background A2A agent indicator
## Summary

The live indicator for in-flight A2A delegations (added in #534) currently
renders a minimal spinner + state line:

    ◓ Agent(browser-agent=working...)

This is enough to know *a* task is running but leaves two questions
unanswered while you wait:

1. **How long has it been running?** A stuck/slow remote agent looks
   identical to a fast one until it terminates. No way to tell at a
   glance whether to keep waiting or cancel.
2. **Which model is the remote agent using?** For cost-sensitive
   delegations (e.g. browser-agent running a paid frontier model vs. a
   cheap flash model) the local CLI has no visibility into what's
   actually executing remotely.

Desired format while non-terminal:

    ◓ Agent(browser-agent=working..., model=deepseek/deepseek-v4-flash) 17s

Terminal states should freeze the elapsed counter and surface the total
duration alongside the existing `usage=` / `execution_stats=` branches:

    ✓ Agent(browser-agent=completed) 42s
      ├── usage={...}
      └── execution_stats={...}

Implementation touch points (for orientation, not a binding design):

- `BackgroundTaskDisplay` in `internal/ui/components/conversation_view.go`
  gains `StartedAt time.Time` and `Model string` fields.
- `StartedAt` is set on `A2ATaskSubmittedEvent` handling (around
  `conversation_view.go:1185`).
- `renderBackgroundTaskLine` and `renderTerminalMultiLine` format
  elapsed seconds; the counter freezes at the moment of terminal
  transition (do not keep ticking).
- A 1s ticker (or reuse the existing `backgroundSpinner` tick if its
  cadence is fast enough) re-renders while any non-terminal task is
  in flight. Stop the ticker when the map is empty to avoid idle work.
- Model source: open question. Candidates:
    1. Remote `Task.metadata` (preferred if ADK exposes it; aligns with
       how `usage` / `execution_stats` are already surfaced).
    2. Agent card discovered at submit time (one-shot, cached per URL).
    3. `~/.infer/agents.yaml` configured default per agent.
  Pick whichever is cheapest to obtain authoritatively; gracefully omit
  the `model=...` segment when the value is unknown so older agents
  still render cleanly.

UX notes:

- Long model identifiers (`provider/some-very-long-model-name`) shouldn't
  blow up the sticky bar width. Consider truncating with an ellipsis if
  the indicator would exceed the terminal width.
- Elapsed format: plain `17s` up to 60s, then `1m23s`. No sub-second
  precision (ticker is 1s).

## Acceptance Criteria

- [ ] `BackgroundTaskDisplay` carries `StartedAt time.Time` and
      `Model string`; both are populated by the time the first frame
      renders after `A2ATaskSubmittedEvent`.
- [ ] While the task is in any non-terminal state, the indicator
      includes the elapsed seconds since submission and re-renders at
      least once per second.
- [ ] When the model is known, the indicator renders
      `Agent(<name>=<state>..., model=<model>) <elapsed>` exactly in
      that order; when unknown, the `model=...` segment is omitted with
      no trailing comma artefact.
- [ ] On terminal transition (completed / failed / cancelled), the
      elapsed counter freezes at the duration between submission and
      that transition, and is surfaced in the multi-line block.
- [ ] Per-second re-render ticker stops when no background tasks are
      tracked (no idle CPU when the map is empty).
- [ ] Indicator gracefully degrades on narrow terminals: long model
      names are truncated rather than wrapping the sticky bar.
- [ ] Tests in `internal/ui/components/conversation_view_test.go` cover:
      the new format with and without a known model, elapsed-time
      freeze on terminal state, and the no-model fallback for older
      agents.

edenreich:
## Summary

- Extends the inline sticky-bar A2A indicator (added in #534) with **elapsed seconds** since submission and the **remote agent's configured model**, addressing #536.
- Non-terminal line: `◓ Agent(browser-agent=working..., model=deepseek/deepseek-v4-flash) 17s`. Terminal line freezes elapsed at `CompletedAt-StartedAt` and omits `model=`: `✓ Agent(browser-agent=completed) 42s`.
- Model source is `~/.infer/agents.yaml` (`AgentEntry.Model` already exists for spawning OCI agents); wired via a new `SetAgentModelResolver` that mirrors the existing name resolver.
- Re-render cadence reuses the existing 100ms spinner tick — no new ticker. The chain self-terminates when the task map empties.
- Long model identifiers truncate with `...`; name, state, and elapsed are preserved. Width is now threaded from `RenderBackgroundTasksBar` through `renderBackgroundTaskLine`.

Closes #536

### DIFF
--- a/internal/ui/components/conversation_view.go
+++ b/internal/ui/components/conversation_view.go
@@ -1513,7 +1513,7 @@ func (cv *ConversationView) formatNonTerminalBody(name, state string, display *B
 	if width <= 0 || model == "" {
 		return body
 	}
-	const iconBudget = 2 // icon + space prepended by caller
+	const iconBudget = 2
 	avail := width - iconBudget
 	if len(body) <= avail {
 		return body
--- a/internal/ui/components/conversation_view.go
+++ b/internal/ui/components/conversation_view.go
@@ -47,12 +47,14 @@ type BackgroundTaskDisplay struct {
 	TaskID             string
 	AgentName          string
 	AgentURL           string
+	Model              string
 	State              string
 	Message            string
 	UsageJSON          string
 	ExecutionStatsJSON string
 	ErrorMsg           string
 	IsTerminal         bool
+	StartedAt          time.Time
 	CompletedAt        time.Time
 }
 
@@ -113,6 +115,10 @@ type ConversationView struct {
 	// from ~/.infer/agents.yaml. Optional; falls back to the URL when nil
 	// or when the URL has no matching entry.
 	agentNameResolver func(url string) string
+	// agentModelResolver maps an agent URL to its configured model (e.g.
+	// "deepseek/deepseek-v4-flash") from ~/.infer/agents.yaml. Optional;
+	// when nil or no match, the model segment is omitted from the indicator.
+	agentModelResolver func(url string) string
 }
 
 func NewConversationView(styleProvider *styles.Provider) *ConversationView {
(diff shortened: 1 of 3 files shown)


internal/ui/components/conversation_view.go
```diff
@@ -1423,6 +1468,78 @@ func (cv *ConversationView) agentDisplayName(display *BackgroundTaskDisplay) str
...
+	case elapsed != "":
+		body = fmt.Sprintf("Agent(%s=%s...) %s", name, state, elapsed)
+	default:
+		body = fmt.Sprintf("Agent(%s=%s...)", name, state)
+	}
+
+	if width <= 0 || model == "" {
+		return body
+	}
+	const iconBudget = 2 // icon + space prepended by caller
```
edenreich:
```suggestion
	const iconBudget = 2
```

PR #546, opened 2026-05-28
feat: observability reminders and compaction

edenreich:
Add extra observability to reminders and auto compaction.

### DIFF
--- a/internal/streamevent/streamevent.go
+++ b/internal/streamevent/streamevent.go
@@ -1,0 +1,75 @@
+// Package streamevent emits structured JSON-line events on stdout
+// describing internal agent lifecycle moments that are otherwise invisible
+// to outside observers - system reminder injection, conversation
+// compaction triggers, and similar checkpoints. Downstream tooling (the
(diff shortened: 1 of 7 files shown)


internal/streamevent/streamevent.go
```diff
@@ -0,0 +1,160 @@
...
+	zapcore "go.uber.org/zap/zapcore"
+
+	logger "github.com/inference-gateway/cli/internal/logger"
+)
+
+var (
+	writerMu sync.Mutex
+	writer   io.Writer = os.Stdout
+
+	// gateMu protects the optional test override.
```
edenreich:
```suggestion
```

internal/streamevent/streamevent.go
```diff
@@ -0,0 +1,160 @@
...
+	gateMu       sync.RWMutex
+	gateOverride *bool // nil = use logger debug level; non-nil = forced value for tests
```
edenreich:
```suggestion
	gateOverride *bool
```

PR #550, opened 2026-05-29
feat(agent): re-check session rollover inside the turn loop (#549)

edenreich:
Closes #549.

## Summary

- `infer agent` previously evaluated `SessionRolloverManager.ShouldRollover` only at session start (`cmd/agent.go:195`). A long agent run could cross both the token-fill (`compact.auto_at`) and idle (`compact.rollover_on_idle_minutes`) thresholds without ever producing a new session file with a summary — the persisted conversation grew monotonically across e.g. a 50-turn run.
- Chat mode already re-evaluates rollover before every user message at `internal/handlers/chat_message_processor.go:239-248`. This PR brings agent mode in line by re-evaluating rollover at the top of every turn-loop iteration.
- The trigger logic that chat and agent now share is lifted onto `*services.SessionRolloverManager` as a new `MaybeRollover(ctx, model, groupKey) (newID string, fired bool)` method so both modes go through the same `ShouldRollover → PerformRollover → warn-on-fail` path.

## Notable design decisions

- **`s.completedTurns` is preserved across rollover.** `--max-turns N` is a user-set hard ceiling; resetting on every rollover would silently extend it and also disrupt the cadence of `injectSystemReminderIfDue`. Documented in the doc comment on `maybeRollover`.
- **The in-loop call sits *before* `injectSystemReminderIfDue`.** A rollover swaps `s.conversation`; if the reminder were injected first, it would be summarised away by the rollover that immediately follows. Placing the rollover first means the next reminder cycle lands cleanly in the new conversation.
- **The idle trigger is structurally a no-op inside an active loop** because every preceding `executeTurn` appends a fresh-timestamped message — the asymmetry is documented near the call site rather than guarded with dead-code branching, per the issue's "confirm and document the asymmetry, or guard explicitly" criterion.

## Code changes

| File | Change |
| --- | --- |
| `internal/services/session_rollover_manager.go` | New `MaybeRollover` method |
| `internal/handlers/chat_message_processor.go` | Consume `MaybeRollover`; behavior unchanged |
| `cmd/agent.go` | Two new `AgentSession` fields (`rolloverManager`, `groupKey`); `maybeRolloverSession` package function becomes `(*AgentSession).maybeRollover`; new `s.maybeRollover()` call at the top of the turn-loop body |
| `cmd/agent_test.go` | `TestMaybeRolloverInLoop` — four cases covering the issue's acceptance criteria |
| `internal/services/session_rollover_manager_test.go` | Four cases for the new `MaybeRollover` method |

## Known limitations (out of scope for #549, mentioned for follow-up)

- **`BackgroundTasksWaiter` retains the pre-rollover session id.** Constructed once at `cmd/agent.go:181-187` with the original `newSessionID`. After mid-loop rollover, in-flight A2A and background-bash tasks continue to be tracked under the pre-rollover id. Will file a follow-up issue if needed.
- **Async persistence race window.** `addMessage` at `cmd/agent.go:993-1005` kicks off a goroutine for `conversationRepo.AddMessage`. If a goroutine from turn N is still in-flight when the loop iterates back into `maybeRollover` and that rollover calls `repo.StartNewConversation`, the late-arriving message may land in the new conversation file. This race exists pre-change for `/compact` flows too; no new exposure.

## Docs follow-up

Per the org's docs follow-up rule: this is a public-surface change (new `MaybeRollover` method + new agent runtime behavior). A `[DOCS]` ticket for `inference-gateway/docs` will reference this PR.

### DIFF
--- a/cmd/agent_test.go
+++ b/cmd/agent_test.go
@@ -1141,9 +1141,6 @@ func TestMaybeRolloverInLoop(t *testing.T) {
 		}
 		originalID := repo.GetCurrentConversationID()
 
-		// Hidden:true messages cross the ShouldRollover gate (via the
-		// gateway-reported LastInputTokens) but get filtered by PerformRollover,
-		// driving the "no visible messages to rollover" error path.
 		if err := repo.AddMessage(domain.ConversationEntry{
 			Message: sdk.Message{Role: sdk.User, Content: sdk.NewMessageContent("hidden")},
 			Time:    time.Now(),
--- a/cmd/agent_test.go
+++ b/cmd/agent_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"io"
@@ -16,6 +17,8 @@ import (
 
 	config "github.com/inference-gateway/cli/config"
 	domain "github.com/inference-gateway/cli/internal/domain"
+	storage "github.com/inference-gateway/cli/internal/infra/storage"
+	services "github.com/inference-gateway/cli/internal/services"
 	streamevent "github.com/inference-gateway/cli/internal/streamevent"
 )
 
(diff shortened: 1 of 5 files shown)


internal/services/session_rollover_manager_test.go
```diff
@@ -395,3 +395,111 @@ func TestConcurrentRolloversForDifferentGroups(t *testing.T) {
...
+		t.Errorf("nil receiver must return empty newID, got %q", newID)
+	}
+}
+
+func TestMaybeRollover_GateClosedReturnsFalse(t *testing.T) {
+	mgr, _, opt, _, cleanup := newRolloverManagerForTest(t, 80, 30)
+	defer cleanup()
+
+	// Empty conversation, idle threshold not crossed, no token usage recorded.
+	// ShouldRollover returns false, so MaybeRollover must short-circuit.
```
edenreich:
```suggestion
```

internal/services/session_rollover_manager_test.go
```diff
@@ -395,3 +395,111 @@ func TestConcurrentRolloversForDifferentGroups(t *testing.T) {
...
+
+	if err := repo.StartNewConversation("Initial"); err != nil {
+		t.Fatalf("start: %v", err)
+	}
+	originalID := repo.GetCurrentConversationID()
+
+	// Hidden:true entries pass the ShouldRollover gate (the trigger inspects
+	// entry count + token usage, not Hidden) but PerformRollover filters them
+	// out and returns "no visible messages to rollover". Exercising this path
+	// verifies MaybeRollover swallows the error and returns ("", false).
```
edenreich:
```suggestion
```

cmd/agent_test.go
```diff
@@ -977,3 +980,193 @@ func TestInjectSystemReminderIfDue(t *testing.T) {
...
+		defer cleanup()
+
+		if err := repo.StartNewConversation("Initial"); err != nil {
+			t.Fatalf("start: %v", err)
+		}
+		originalID := repo.GetCurrentConversationID()
+
+		// Hidden:true messages cross the ShouldRollover gate (via the
+		// gateway-reported LastInputTokens) but get filtered by PerformRollover,
+		// driving the "no visible messages to rollover" error path.
```
```suggestion
```

PR #572, opened 2026-06-01
refactor: remove built-in Github tool in favor of the gh CLI

edenreich:
## Summary

Closes #570. Removes the built-in `Github` tool and standardizes on the `gh` CLI (via Bash). The built-in tool multiplexed 8 operations behind one `resource` param (models routinely omitted required fields like `issue_number`/`head`), authenticated from a separate `GITHUB_TOKEN` instead of the `gh` credential chain (403 "Resource not accessible by integration" in CI), and duplicated `gh` with worse error messages.

**Breaking change** (`feat!`): the `tools.github` config block and the `infer config tools github` commands are removed. Configs that still contain a `tools.github` section are ignored (viper drops unknown keys), so they don't error.

## Changes

- **Removed**: `internal/agent/tools/github.go` (+ tests), its registration, the `tools.github` config (`GithubToolConfig`/`GithubSafetyConfig`, defaults, approval case), the `infer config tools github` command subtree, the GitHub API domain types (`GitHubIssue`/`GitHubPullRequest`/`GitHubComment`/…), and the now-orphaned `DetectGithubOwner` helper. The separate **GitHub Action setup wizard** is untouched.
- **Whitelist** (`config/config.go` + `.infer/config.yaml`): non-destructive `gh` ops are auto-approved by default — read-only `gh <noun> list|view|status|diff|checks`, `gh auth status`, `gh issue create|edit|comment`, `gh pr create`, and **GET-only** `gh api` (bare endpoint + read-only flags). Plus `env`/`printenv`. Destructive ops (`gh pr merge`, `gh repo delete`, mutating `gh api`, …) are deliberately omitted so they fall through to the approval flow — not blocked.
- **Prompt steering**: a dynamic `GITHUB OPERATIONS` context block (reaches existing users regardless of their `prompts.yaml`, gated on Bash being enabled) plus a sentence in the default system prompt, both pointing at `gh`.
- **`infer init`**: the seeded PR-creation snippet now uses `gh pr create` instead of the removed tool.

### Note on `gh api` GET-only

Go's RE2 has no negative lookahead, so "`gh api` but not `-X POST`" can't be expressed directly. The pattern `^gh api [^ -][^ ]*( --paginate| --jq <expr>| -q <expr>)*$` auto-approves a bare endpoint with only read-only flags; anything with `-X`/`--method`/`-f`/`-F`/`--field`/`--input` (writes) falls through to approval. Safe by design.

## Tests

- `TestIsBashCommandWhitelisted_GhDefaults`: asserts each read + targeted-write is whitelisted and the destructive set (`gh pr merge`, `gh repo delete`, `gh api … -X POST`, `--method POST`, `-f …`, leading-flag api) is **not**.
- Existing `config`/`prompts` tests updated; gate grep for `Tools.Github|GithubToolConfig|NewGithubTool|GitHubIssue|…|tools.github` comes back empty.
- `task test` + `task lint` green. Smoke-tested the built binary: `infer config tools github` no longer exists; `infer config tools validate` confirms the gh allow/deny split.

## Follow-up

- A matching `[DOCS]` ticket for `inference-gateway/docs` (removal of the GitHub tool + `gh` whitelist guidance) should accompany this PR, per the maintainer docs rule.


config/config.go
```diff
@@ -668,6 +650,19 @@ func DefaultConfig() *Config { //nolint:funlen
...
 						"^git remote( -v)?$",
 						"^git show",
+						// gh CLI: non-destructive operations. Reads (list/view/
+						// status/diff/checks), the targeted writes the agent
+						// needs (issue create/edit/comment, pr create), and
+						// GET-only `gh api` (bare endpoint + read-only flags).
+						// Destructive ops (pr merge, repo delete, release
+						// create/delete, run cancel, auth login, workflow run,
+						// `gh api -X POST/...`) are deliberately omitted so they
+						// fall through to the approval flow.
```
edenreich:
```suggestion
```

PR #605, opened 2026-06-05
fix(tools): make bash whitelist shell-aware to fix benign-command rejections (#581)

edenreich:
## Summary

Fixes #581. An `infer` session reported a 36% tool-call success rate because benign, read-only GitHub/git commands were rejected by the bash whitelist — mostly because the model appended `2>&1` / `2>/dev/null` / `|| echo ...`, which broke the (often end-anchored) whitelist patterns. The same naive prefix/regex matching also let a whitelisted *head* command carry an arbitrary tail, so the boundary was both **too strict for benign commands and too loose to be a real boundary**.

This makes the matcher *shell-aware* and unifies the previously-duplicated logic (`bash.go::isCommandAllowed` and `config.go::IsBashCommandWhitelisted`) into a single source of truth — `config.IsBashCommandWhitelisted` — now consulted by the Bash tool, the approval policy, and agent auto-approval, so all three agree on exactly what runs without prompting.

## What changed

**Reliability (the reported failures):**
- **Strip benign trailing redirections** (`2>&1`, `2>/dev/null`, `&>/dev/null`, fd-dups) per segment before matching, so they no longer defeat end-anchored patterns. → fixes `gh api … 2>&1`, `git push … 2>&1`, etc.
- **`gh api --jq`/`-q` now accept quoted, space-containing expressions** — e.g. `gh api repos/o/r --jq '.[] | .id'`, the canonical read idiom (previously rejected because `--jq [^ ]+` forbade spaces).
- **Add read-only `gh search (issues|code|prs|repos|commits)`**.
- **`git status` flags allowed** (`git status --porcelain`) — aligns it with the already-unanchored `git log`/`git diff`/`git show` read patterns.
- **System-prompt guidance**: the Bash tool already captures stdout/stderr/exit code, so don't append `2>&1`, `2>/dev/null`, or `|| echo ...`; prefer `gh api` over fetching `api.github.com` directly. Lives in the dynamic context, so it reaches existing users regardless of their `prompts.yaml` override.

**Security coherence (the "no real boundary" critique):**
- **Compound commands** (`&&`, `||`, `|`, `|&`, `;`, `&`, newlines) are split **quote-aware**, and every segment must be independently whitelisted. Enables benign chains (`gh issue view 5 && gh issue comment 5 …`) while closing the prefix-match hole (`gh issue list && rm -rf /` is now rejected).
- **Reject command substitution**: `$(...)`, backticks, and process substitution `<(...)`/`>(...)` (e.g. `echo $(rm -rf /)`).
- **Reject dangerous `find` primaries** that execute or delete: `-exec`, `-execdir`, `-ok`, `-okdir`, `-delete`, `-fprintf`, … (read-only `find -name`/`-type` is unaffected).

## Verification

- `task test` green; `golangci-lint` reports **0 issues**; markdownlint clean.
- New exhaustive matcher tests cover redirect stripping, compound chains, command substitution, quoted operators, unbalanced quotes, `gh search`, `gh api --jq`, `find` actions, and `git status` flags; plus tokenizer unit tests and a Bash-tool integration test.
- Ran a **6-lens adversarial security review** (multi-agent) and executed every candidate command through the real compiled matcher for ground truth. It reduced confirmed issues from 27 → 17; the remaining 17 are all **pre-existing** and unrelated to the reliability fix.

## Deliberately out of scope (pre-existing; documented, not bundled)

These were surfaced by the adversarial review but **not** changed here because fixing them carries real false-reject cost (the opposite of this issue's goal) or belongs in a separate hardening decision:

- **Bare `$VAR` expansion** (`echo $GITHUB_TOKEN`) — blocking `$` would reject `echo $HOME`, `git log --since=$DATE`, etc.
- **Redirects to real files** (`echo … > /tmp/x`) — blocking would reject legitimate `gh pr diff > out.patch`. The redirect just changes where already-permitted output goes.
- **`git branch` flag loosening** — risky because `-D`/`-m` mutate.

Happy to file a focused follow-up issue for `$VAR`/file-redirect hardening if desired.

> Note: arithmetic expansion (`echo $((1+1))`) is conservatively treated like command substitution and falls through to approval. Negligible in practice.

### DIFF
--- a/config/bash_whitelist_test.go
+++ b/config/bash_whitelist_test.go
@@ -119,8 +119,6 @@ func TestIsBashCommandWhitelisted_CommandSubstitution(t *testing.T) {
 		}
 	}
 
-	// Single-quoted substitution syntax is a literal string in bash, so it is
-	// safe and the wrapping command is judged on its own.
 	allowed := []string{
 		"echo '$(rm -rf /)'",
 		"gh issue create --body 'use $(x) verbatim'",
@@ -280,8 +278,8 @@ func TestStripBenignTrailingRedirections(t *testing.T) {
 		{"echo hi >&2", "echo hi"},
 		{"echo hi 2>&-", "echo hi"},
 		{"echo hi", "echo hi"},
-		{"echo /dev/null", "echo /dev/null"},                   // an arg, not a redirect
-		{"gh issue list > out.txt", "gh issue list > out.txt"}, // real file: kept
+		{"echo /dev/null", "echo /dev/null"},
+		{"gh issue list > out.txt", "gh issue list > out.txt"},
 		{"2>&1", ""},
 	}
 	for _, tt := range tests {
--- a/config/bash_whitelist_test.go
+++ b/config/bash_whitelist_test.go
@@ -1,0 +1,362 @@
+package config
+
+import (
+	"strings"
+	"testing"
+)
+
+// pushCfg builds a config whose only whitelist entries are the end-anchored
+// git-push patterns, mirroring what users configure for branch protection. It
+// exercises the interaction between redirection stripping and "$"-anchored
+// patterns (a redirect suffix must not defeat the anchor).
+func pushCfg() *Config {
+	return &Config{
+		Tools: ToolsConfig{
+			Enabled: true,
+			Bash: BashToolConfig{
+				Enabled: true,
+				Whitelist: ToolWhitelistConfig{
+					Patterns: []string{
+						"^git push( --set-upstream)?( origin)? (feature|fix)/[a-zA-Z0-9/_.-]+$",
+					},
+				},
+			},
+		},
+	}
+}
+
+func TestIsBashCommandWhitelisted_RedirectStripping(t *testing.T) {
+	cfg := DefaultConfig()
+
+	allowed := []string{
+		"gh api repos/o/r/issues 2>&1",
+		"gh issue list 2>/dev/null",
+		"gh issue list >/dev/null 2>&1",
+		"gh issue list 2>&1 >/dev/null",
+		"gh pr view 5 2>&1",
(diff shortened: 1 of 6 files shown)


config/bash_whitelist_test.go
```diff
@@ -0,0 +1,362 @@
...
+		`echo "leak: $(printenv)"`,
+	}
+	for _, cmd := range denied {
+		if cfg.IsBashCommandWhitelisted(cmd) {
+			t.Errorf("expected %q NOT to be whitelisted (contains command substitution)", cmd)
+		}
+	}
+
+	// Single-quoted substitution syntax is a literal string in bash, so it is
+	// safe and the wrapping command is judged on its own.
```
edenreich:
```suggestion
```

config/bash_whitelist_test.go
```diff
@@ -0,0 +1,362 @@
...
+		{"gh api repos/o/r 2>&1", "gh api repos/o/r"},
+		{"gh issue list 2>/dev/null", "gh issue list"},
+		{"gh issue list >/dev/null 2>&1", "gh issue list"},
+		{"gh issue list 2>&1 >/dev/null", "gh issue list"},
+		{"gh issue list &>/dev/null", "gh issue list"},
+		{"echo hi >&2", "echo hi"},
+		{"echo hi 2>&-", "echo hi"},
+		{"echo hi", "echo hi"},
+		{"echo /dev/null", "echo /dev/null"},                   // an arg, not a redirect
+		{"gh issue list > out.txt", "gh issue list > out.txt"}, // real file: kept
```
edenreich:
```suggestion
		{"echo /dev/null", "echo /dev/null"},
		{"gh issue list > out.txt", "gh issue list > out.txt"},
```

PR #610, opened 2026-06-06
feat: add npm wrapper to install and run the CLI via npx

edenreich:
## Summary

Closes #568. Adds an npm/npx wrapper so the CLI can be run with `npx @inference-gateway/cli` (Node >= 18) - no Go toolchain or manual download. Mirrors the approach already shipping in `inference-gateway/adl-cli`.

This repo ships raw release binaries (`infer-<os>-<arch>`) built by `artifacts.yml`, so the installer downloads the binary directly (no archive to extract) and verifies it against `checksums.txt`.

## What's included

- **`npm/` wrapper (zero runtime deps)**
  - `package.json` - `@inference-gateway/cli`, bin `infer`, `postinstall` hook, `os`/`cpu` gates (linux/darwin, x64/arm64).
  - `bin/install.js` - downloads the matching `infer-<os>-<arch>` release asset, checksum-verifies against `checksums.txt`, caches it next to the script, `chmod +x`. Honors `INFER_CLI_BASE_URL` and `INFER_CLI_SKIP_DOWNLOAD`.
  - `bin/run.js` - locates/installs the binary, execs it, forwards args and exit code.
  - `.gitignore` - excludes the fetched binary and the publish-time README/LICENSE copies.
- **`.releaserc.yaml`** - `@semantic-release/npm` (`pkgRoot: npm`); copies LICENSE/README into `npm/` at release; commits `npm/package.json` so the version stays in sync; adds an "npm / npx" block to the release body.
- **`.github/workflows/release.yml`** - `id-token: write` (OIDC), Node pinned to `24` (npm >= 11.5.1), `@semantic-release/npm@13.1.5`, passes `NPM_TOKEN` for the bootstrap publish.
- **`README.md`** - new "Using npm/npx (Recommended)" install section with a "not recommended for production" caveat.
- **`Taskfile.yml` + `.markdownlintignore`** - moved markdownlint ignores into a config file so they apply for `task lint`, pre-commit, and any direct/CI invocation; the `npm/` publish copies are excluded.

`artifacts.yml` is unchanged - it already produces `infer-{linux,darwin}-{amd64,arm64}` + `checksums.txt`, exactly what the installer consumes.

## Platforms

Linux + macOS on amd64/arm64 (matches adl-cli and this repo's release assets). armv7 and Windows (use WSL) are out of scope.

## Manual prerequisites before the next release publishes to npm (org admin)

1. **Bootstrap:** add a temporary `NPM_TOKEN` secret (npm automation token with publish rights to the `@inference-gateway` scope). The first release publishes `@inference-gateway/cli` using it.
2. **Switch to OIDC:** after the first publish, configure a trusted publisher on npmjs.com for `@inference-gateway/cli` -> repo `inference-gateway/cli`, workflow `.github/workflows/release.yml`, then delete the `NPM_TOKEN` secret so subsequent releases publish via OIDC (same model as `@inference-gateway/adl-cli`).

## Verification

- `install.js` downloaded + checksum-verified `infer-darwin-arm64` from v0.118.0; `run.js` ran it (`infer v0.118.0`, `--help`); re-run no-ops (cached); `INFER_CLI_SKIP_DOWNLOAD` skips.
- `npm pack --dry-run` -> tarball = package.json + bin/{run,install}.js + README + LICENSE (native binary excluded).
- `markdownlint .` and the full pre-commit lint pass.

## Follow-up

Docs ticket filed: inference-gateway/docs#182 (document `npx @inference-gateway/cli`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v6.4.0
         with:
-          node-version: '24'
+          node-version: '24.15.0'
 
       - name: Cache npm global packages
         uses: actions/cache@v5.0.5
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,7 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
+      id-token: write
     steps:
       - name: Create GitHub App Token
         uses: actions/create-github-app-token@v3.2.0
@@ -54,7 +55,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v6.4.0
         with:
-          node-version: 'lts/*'
+          node-version: '24'
 
       - name: Cache npm global packages
         uses: actions/cache@v5.0.5
@@ -74,6 +75,7 @@ jobs:
             @semantic-release/exec@7.1.0 \
             @semantic-release/git@10.0.1 \
             @semantic-release/github@1.0.0 \
+            @semantic-release/npm@13.1.5 \
             conventional-changelog-conventionalcommits@9.1.0
 
       - name: Check for existing releases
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -54,7 +55,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v6.4.0
         with:
-          node-version: 'lts/*'
+          node-version: '24'
```
edenreich:
```suggestion
          node-version: '24.15.0'
```

PR #618, opened 2026-06-06
feat(config): single-command bash whitelist + env-var leak guard

edenreich:
## Summary

This branch began as a hardening of what the Bash tool auto-approves and grew into a broader pass over **agent safety**, **headless secure-by-default behaviour**, and a batch of **chat-TUI / diff-viewer UX** improvements. The sections below cover everything that ships in this PR.

> Note: the bash auto-approval vocabulary was reworked **within this branch** from the old `whitelist` model into a per-mode **allowed-list**. The descriptions below reflect the final shipped state (`tools.bash.mode.<mode>.allow`, `INFER_TOOLS_BASH_ALLOW_APPEND`) — the earlier `INFER_TOOLS_BASH_WHITELIST_COMMANDS*` names no longer exist.

---

## 1. Bash auto-approval: per-mode allowed-list (default-deny)

Reworks bash auto-approval into a single, config-driven, **per-mode allowed-list**. A command auto-runs only if it matches the allowed-list for the active agent mode; anything unmatched is denied — an approval prompt in chat, a hard reject with an actionable hint in headless agent mode. There is no separate deny list.

- **Per-mode config** — `tools.bash.mode.{all,plan,standard,auto}.allow`. `mode.all` is the every-mode baseline, unioned with the active mode's own list (`bashAllowFor`). `mode.auto` defaults to the `.*` sentinel (unrestricted); `mode.plan` / `mode.standard` add nothing and reduce to the read-only baseline.
- **Full-command matching** — entries match `\A(?:entry)\z`, so a bare token like `gh` allows only `gh` (never `gh issue list`); entries opt into arguments explicitly (`gh issue.*`). The single `.*` sentinel means *unrestricted* and skips the clean-command guard.
- **Single matcher** — `IsBashCommandAllowed(command, mode)` is shared by the Bash tool gate, the approval policy, and agent auto-approval. Mode flows in via `domain.WithAgentMode` context + `AgentMode.AllowedlistKey()`.
- **Clean-command guard** (every non-`.*` mode) rejects, before matching:
  - command substitution (`$(…)`, backticks, `<()`/`>()`),
  - multi-command chains / pipelines (top-level `|`, `&&`, `||`, `;`, `&`, newline — operators inside quotes don't count, so `jq '… | …'` stays a single command). This closes the `echo x | xargs rm` prefix hole.
  - a surviving file-write redirect (`>`/`>>`; benign `2>&1` / `>/dev/null` are stripped first),
  - dangerous `find` actions (`-exec`/`-delete`/…),
  - the **env-var leak guard**: a printing/publishing command (`echo`/`printf`/`gh issue|pr create|comment|edit`) may not expand `$VAR`, so `echo $AWS_SECRET_ACCESS_KEY` is blocked while `ls $DIR` stays allowed (a single-quoted or escaped `$` is literal).
- **Hardened defaults** — the raw `gh api` GET wildcard is dropped; the baseline now enumerates explicit non-destructive `gh` subcommands (`issue|pr|repo|release|run|workflow list|view|…`, `gh search`, and read-only `gh project list|view|item-list|field-list`). `gh project` *writes* (`item-add`/`item-edit`), `git push`, and `git commit` are **not** in the defaults and fall through to approval. A raw-`gh api` need is opt-in per repo.
- **Append-only override** — `--tools-bash-allow-append` / `INFER_TOOLS_BASH_ALLOW_APPEND` (comma/newline list, env wins over flag) merges onto `mode.all.allow` after config load, so CI can add a few safe commands (e.g. `git commit` / `git push`) without rewriting config or shipping `.*`.
- **Model feedback** — each guard rejection returns an actionable hint (`BashCommandRejectionHint`), and the effective per-mode allow-list is injected into the system prompt (`buildBashAllowInfo`), rebuilt each turn so a chat mode toggle re-injects it. The model is told not to retry a denied command but to stop and ask, or use an allowed alternative.

## 2. Headless secure-by-default (`approval_behaviour`)

Headless `infer agent` previously defaulted to auto mode (`.*`), so CI, heartbeat, and any run without `--require-approval` could execute **any** command unattended. This makes the restricted path the default.

- New `tools.safety.approval_behaviour` (`prompt` | `ipc` | `block`, default `prompt`) decides **how** a needed approval is delivered, alongside the existing `require_approval` which decides **whether**. `config.ResolveApprovalDelivery(behaviour, brokerAttached, isChat)` is the single resolver; `Config.Validate` rejects unknown values.
- The headless executor now runs in **standard** mode and routes approval-requiring tools per behaviour: **IPC** when a broker is attached (`--require-approval`, e.g. the channel manager), otherwise **blocked** with a reason. No more `.*` default — full autonomy is now an explicit opt-in.
- Chat honours `block` (rejects without prompting); `prompt` / `ipc` still prompt.
- **Auto-accept safety prompt** — `prompts.agent.system_prompt_auto` carries a destructive-action policy (confirm/avoid irreversible ops: delete, force-push, drop, `rm -rf`, publish) for auto-accept mode, wired via `getSystemPromptForMode`; falls back to `system_prompt` when blank.
- **Controlled-autonomy CI profile**: `approval_behaviour: block` + `tools.write.require_approval: false` + a curated bash allow-list / the append override.

## 3. `BlockingTools` agent state for undeliverable approval

A new `StateBlockingTools` handles the case where a tool needs approval but no approver is reachable (`approval_behaviour=block`, or `=ipc` with no broker).

- Entered from `EvaluatingTools` when approval is required **and** undeliverable for all gated tools in the batch.
- Walks the batch in order: non-gated tools (read-only, allow-listed Bash) still execute; gated tools are rejected with an actionable reason telling the model not to retry but to explain what it needs. Results are appended in conversation order, mirroring the `ApprovingTools` contract.
- Emits a terminal tool-progress event so the live streaming overlay clears instead of lingering, then transitions to `PostToolExecution`.

## 4. Tool-execution stop routed through the event loop (fix)

The `executeTools` goroutine transitioned to `StateStopped` directly, but `ExecutingTools → Stopped` was never registered and the goroutine emitted no event, so `processEvents` never woke to observe the terminal state — the loop hung until session cancel (reachable via a rejected tool result or a successful `RequestPlanApproval`).

- `executeTools` now only emits `ToolsCompletedEvent{Stop}`; the `ExecutingTools` state owns the transition (`Stop → Stopped`). Registers `ExecutingTools → Stopped` as the single needed from-state.
- `Completing` now explicitly consumes `CompletionRequestedEvent` rather than finalizing on any event.
- State journey is now debuggable: the transition chokepoint log is bumped to Info (file-only) and a Debug `dispatching event` line carries event + state + request_id.

## 5. Tool approval UX: diff preview

When approval is required for file-mutating tools (**Edit / MultiEdit / Write**), the approval box now renders a **colored, syntax-aware diff preview** (file path + old/new content, plus a "replace all occurrences" note for `replace_all`) instead of the compact one-liner. The preview is height-capped (≈ half the terminal, bounded 6–30 lines) with a dim "… N more lines (full diff shown after approval)" hint when truncated. All other tools keep the compact `Name(arg=value, …)` summary.

## 6. Agent skills enabled by default

`AgentSkillsConfig.Enabled` now defaults to `true`, so skills are discovered and injected into the system prompt out of the box. CLI help text, tests, and the seeded `config.yaml` are updated accordingly.

---

## Chat-TUI / diff-viewer / git-staging UX

- **Patch line selection + hunk split** — in the diff viewer you can select a line range within a hunk and stage/unstage only those lines (default: `space`/`v` to anchor a selection, navigate, apply with the stage/unstage keys), and split a multi-change hunk into independent blocks (`s`) so each can be staged separately. The hint text reflects whether a selection is active.
- **Stage all / unstage all** — stage every change at once (default `A` → `git add -A`, incl. untracked & deletions) or unstage everything (default `U` → `git reset -q HEAD`), from anywhere in the panel.
- **Staged-file hint** — the footer hides the `discard` option when a staged file is selected, since discard only applies to unstaged changes.
- **Removed file-type emoji icons** from the diff renderer; file paths render plain.
- **Clipboard status flashes** — copy/paste now show brief auto-dismissing status messages ("Copied to clipboard", "Text pasted from clipboard", "Image pasted from clipboard") that preserve any active loading spinner.
- **Image placeholder fix** — attachments are labeled `[Image 1]` instead of `[Image #1]`, so the `#` is no longer parsed as a Markdown heading.
- **Unified `/model` command** — `/model <name>` permanently switches the active model (replaces `/switch`), while `/model <name> <prompt…>` runs a single message with that model and restores the session model afterward. `/switch` is deprecated.
- **Autocomplete no longer auto-executes** — `Tab` always completes the suggestion (fills input + trailing space, never submits); `Enter` submits the current input. Prevents accidental command execution on selection.
- **Theme-aware editor background** — the embedded vim/pty editor launches with `set background=dark|light` (and `syntax enable`) matched to the active TUI theme, so editor colors aren't washed out.
- **Bash streaming output coalescing** — the Bash tool now buffers streamed output and flushes at most every ~50 ms or per ~64 KB, so high-volume commands (large `git diff`, logs) no longer flood the TUI with per-line update cycles. Captured output is unaffected.

## Internal cleanup

- Removed 21 never-used chat/agent event types and their orphaned publish sites (4 of which were broadcast to the Go→Swift bridge with no Go consumer and dropped as "ambiguous", which also removes per-transition log noise to the floating window). State-transition observability is retained via the adjacent `logger.Debug`. *(internal-only refactor — no docs ticket.)*
- Lowercased remaining logging statements for consistency; check the scanner error after agent output; consistent named imports for third-party libraries.

---

## Breaking changes / migration

- **`tools.bash.whitelist.commands` is removed.** Configure `tools.bash.mode.<mode>.allow` instead. The `INFER_TOOLS_BASH_WHITELIST_COMMANDS[_APPEND]` env vars and `--tools-bash-whitelist-commands*` flags are gone; use `INFER_TOOLS_BASH_ALLOW_APPEND` / `--tools-bash-allow-append` (append onto the `mode.all` baseline).
- **Headless `infer agent` no longer auto-approves everything.** It runs in standard mode; off-list / mutating actions are blocked in CI/heartbeat and sent for IPC approval under a channel manager. Restore full autonomy explicitly via a curated allow-list, the append override, or `mode.auto`/`.*`.
- **Compound/piped commands and `echo $VAR` no longer auto-approve** — they now require approval. Re-add specific safe commands via the append override.
- **Agent skills are now on by default** (`agent.skills.enabled: true`).

## Verification

`go test ./...`, `golangci-lint run`, `gofmt`, and `markdownlint` all pass. Seeded config files were regenerated (`go run . init --overwrite`) with user-curated files (`agents.yaml`, etc.) restored.

## Docs follow-up (`inference-gateway/docs`)

Filed: inference-gateway/docs#194 — covers the per-mode bash allowed-list (`tools.bash.mode.<mode>.allow`) + `INFER_TOOLS_BASH_ALLOW_APPEND`, `tools.safety.approval_behaviour` and the controlled-autonomy CI profile, agent skills now default-on, the unified `/model` command, and the new diff-viewer keybindings (patch line selection, hunk split, stage/unstage all). It also flags the stale "whitelist" wording still present in `cli.md` and `github-action.md`. Internal refactors (dead-event removal, stop-routing fix) need no docs ticket.


cmd/root_test.go
```diff
@@ -5,78 +5,72 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
```
edenreich:
```suggestion
	assert "github.com/stretchr/testify/assert"
```

cmd/root_test.go
```diff
@@ -88,20 +82,160 @@ func TestBashWhitelistEnvironmentVariables(t *testing.T) {
...
+		"INFER_TOOLS_BASH_WHITELIST_COMMANDS",
+		"INFER_TOOLS_BASH_WHITELIST_COMMANDS_APPEND",
+	}
+	clear := func() {
+		for _, k := range whitelistEnv {
+			assert.NoError(t, os.Unsetenv(k))
+		}
+	}
+	// Resolve against a config-free HOME so the base is the built-in default rather
+	// than a developer's ~/.infer/config.yaml.
```
edenreich:
```suggestion
```

cmd/root_test.go
```diff
@@ -88,20 +82,160 @@ func TestBashWhitelistEnvironmentVariables(t *testing.T) {
...
+	setup := func(t *testing.T) {
+		t.Helper()
+		t.Setenv("HOME", t.TempDir())
+		clear()
+		t.Cleanup(clear)
+	}
+
+	defaults := config.DefaultConfig().Tools.Bash.Whitelist
+	const defaultCommand = "^task( |$)" // present in the default command list (now a regex)
```
edenreich:
```suggestion
	const defaultCommand = "^task( |$)"
```

cmd/root_test.go
```diff
@@ -88,20 +82,160 @@ func TestBashWhitelistEnvironmentVariables(t *testing.T) {
...
+	reset := func() {
+		for _, k := range whitelistEnv {
+			assert.NoError(t, os.Unsetenv(k))
+		}
+		for _, f := range whitelistFlags {
+			assert.NoError(t, rootCmd.PersistentFlags().Set(f, ""))
+		}
+	}
+	// Resolve against a config-free HOME so the base is the built-in default, and
+	// reset the package-global flags before and after each case so they do not bleed.
```
edenreich:
```suggestion
```

cmd/root_test.go
```diff
@@ -88,20 +82,160 @@ func TestBashWhitelistEnvironmentVariables(t *testing.T) {
...
+		reset()
+		t.Cleanup(reset)
+	}
+	setFlag := func(t *testing.T, name, value string) {
+		t.Helper()
+		assert.NoError(t, rootCmd.PersistentFlags().Set(name, value))
+	}
+
+	defaults := config.DefaultConfig().Tools.Bash.Whitelist
+	const defaultCommand = "^task( |$)" // present in the default command list (now a regex)
```
edenreich:
```suggestion
	const defaultCommand = "^task( |$)"
```

PR #643, opened 2026-06-20
ci(release): attach binaries before publish for immutable releases

edenreich:
## What & why

Immutable releases were enabled org-wide (supply-chain hardening). They seal a release the moment it's published — assets can no longer be added/changed, and a release attestation is generated over the assets present at publish time.

Our pipeline uploaded binaries **after** publish, so every release now fails:

```
HTTP 422: Cannot upload assets to an immutable release.
```

`release.yml` published an asset-less release; `artifacts.yml` (on the `release: published` event) then tried `gh release upload` onto the already-sealed release → 422.

## Root cause

`@semantic-release/github` is itself immutable-safe: given an `assets` list it does **create draft → upload assets → publish** atomically — exactly GitHub's required flow. We just weren't using it; binary upload was offloaded to a separate post-publish workflow. The fix is to attach the binaries via semantic-release instead of uploading them after the fact.

## Change

Collapse `release.yml` + `artifacts.yml` into one ordered pipeline:

| Job | Does |
|---|---|
| **version** | `semantic-release --dry-run` → outputs the next version (no publish) |
| **build** | 4-runner matrix (linux Docker + native macOS CGO + Swift), stamped with that version |
| **release** | sign checksums → assets-present guard → `semantic-release` attaches `dist/*` and publishes atomically |
| **container** | push images (after publish; doesn't touch the release) |

- delete `artifacts.yml`
- add `assets` (4 binaries + checksums + sigstore bundle) and a `verifyReleaseCmd` (dry-run version output) to `.releaserc.yaml`
- pin `@semantic-release/github` `1.0.0` → `12.0.8` (the stale pin was shadowed by the bundled v12; corrected for clarity)

## Preserved / no downstream impact

- asset names unchanged (`infer-<os>-<arch>`) → `install.sh`, npm binary fetch, and the release-notes "Binary Download" section keep working
- embedded `v`-prefixed version string, native macOS CGO + Swift build, and container push — unchanged
- the `release: published` event still fires (external consumers unaffected)
- **bonus:** binaries are now covered by the release attestation — the provenance this setting was enabled for

## Verification

- `actionlint` clean; both YAML files parse; no dangling references to the deleted workflow
- Recommended before the stable stream: cut a prerelease from an `rc/*` branch (already a configured prerelease channel) and confirm `gh release view <rc-tag> --json assets` lists all 6 assets, the release is immutable, the attestation is present, and there's no 422 anywhere.

## Org-wide follow-up

The same post-publish `gh release upload` / `softprops/action-gh-release` / goreleaser-upload pattern likely breaks the same way in other `inference-gateway/*` repos now that immutable releases are enforced org-wide. Worth a sweep.


.github/workflows/release.yml
edenreich:
```suggestion
            semantic-release@25.0.5 \
```

Issue #525, opened 2026-05-21
[TASK] Refactor the prompt for remote control via channel manager


edenreich:
## Summary

Currently the default prompt is too verbose resulting in the LLM generating too much output tokens just for a simple "Hi" prompt - it was absorbed using the Telegram bot integration. This was probably necessary when the LLMs were not good before, but since they became better at answering and staying in focus it's probably wiser also to remove the system reminder injection and see how well it performs.

### Acceptance Criteria

- [ ] The system prompt of the remote control agent is simpler
- [ ] The agent reply fast for simple prompts - it's verified
- [ ] The system reminders is disabled by default to avoid unnecessary context injection - LLMs became better and better that they probably don't need it anymore - it's verified
- [ ] Documentation is adjusted with the new defaults (if applicable)
- [ ] Tests are passing (if applicable)

Issue #270, opened 2025-11-27
[FEATURE] Implement Hooks


edenreich:
## Summary

Hooks are things we can execute when the agent finished generating. This is useful when we need deterministic output. Some might need it, some might not, so will keep it as a feature flag.

### Acceptance Criteria

- [ ] There is a way to execute a arbitrary code when the agent finished generating 
- [ ] It's extensible
- [ ] It's a feature flag, turned off by default
- [ ] It's documented
- [ ] It's tested

PR #675, opened 2026-06-28
feat(reminders): configurable reminders on a shared hook-point catalog

edenreich:
## Summary

Generalizes the two hard-coded system reminders (one periodic every N turns, one wrap-up near `max_turns`, both injected only pre-stream — with a duplicate copy in the headless `infer agent` loop) into a **configurable list of named reminders** attached to a **shared, pre-defined catalog of agent-loop hook points**, living in their own dedicated config file.

Closes #669. Establishes the shared catalog + dispatch seam that #270 (Implement Hooks) builds on.

## The model: Hook = point (WHERE) + action (WHAT)

A fixed, validated catalog of 8 symmetric pre/post hook points (`domain.HookPoint`):

| pre | post |
| --- | --- |
| `pre_session` | `post_session` |
| `pre_stream` | `post_stream` |
| `pre_tool` | `post_tool` |
| `pre_queue_drain` | `post_queue_drain` |

- #669 ships the **reminder** action (inject hidden text). #270 will add the **command** action (run code) at the same `dispatchHooks(point)` seam.
- A reminder carries a **trigger** gating which firings it acts on: `always` (default), `interval`, `turns_before_max` (replaces the old wrap-up), `once`.
- Multiple reminders **stack** on one point — you can nudge about memory *and* todos independently.

## What changed

- **`internal/domain/hooks.go`** (new): the `HookPoint` catalog + `SystemReminder` value type.
- **`domain.SystemReminderProvider`** (new interface, counterfeiter-mocked): the agent depends on the interface, not the concrete config.
- **`config/reminders.go`** (new): a **dedicated `reminders.yaml`** config (top-level `RemindersConfig` with its own loader/default/env-override), the `ReminderConfig` type, the pure `RemindersDue(...)` evaluator implementing the provider interface, and validation (`Config.Validate` rejects unknown hooks/triggers).
- **Event-driven agent**: one `dispatchHooks` seam wired at all 8 points across the state executors; a per-run fired-set (mutex-guarded, since pre-stream dispatch runs on a different goroutine than the other points) backs the `once` trigger.
- **Headless `infer agent`**: unified onto the same provider/seam (removing the duplicated reminder logic), wiring the clean-boundary points (`pre_session`, `pre_stream`, `post_tool`, `post_session`).
- **Correctness guard**: a reminder (a user message) is never injected between an assistant turn's `tool_calls` and its results, so the mid-turn hooks can't orphan a tool call.

## Migration (no backward compatibility)

Reminders move out of `prompts.yaml` (`agent.system_reminders`) into a dedicated `.infer/reminders.yaml`, and the flat fields are removed:

```yaml
# before — prompts.yaml
agent:
  system_reminders:
    enabled: false
    interval: 4
    reminder_text: "..."
    wrap_up_text: "..."
    wrap_up_threshold: 0

# after — reminders.yaml
enabled: false
reminders:
  - name: todo-hygiene
    text: "..."
    hook: pre_stream
    trigger: interval
    interval: 4
```

`infer init` seeds `reminders.yaml`; the default ships one disabled `todo-hygiene` reminder (preserving the off-by-default posture from #525). The master switch also takes `INFER_REMINDERS_ENABLED`. Because reminders ship disabled by default, real-world impact is limited to configs that customized those fields. The companion executable hooks (#270) will get their own `hooks.yaml`, keeping "inject text" and "run code" as separate concerns.

## Test plan

- `task build` — passes
- `task test` (full suite) — passes
- `go test -race ./internal/agent/... ./cmd/... ./config/...` — clean (validates the fired-set across goroutines)
- `task lint` — 0 issues
- `config/reminders_test.go`: trigger truth table, hook filtering, stacking, `once` suppression, validation, default config, and a save/load round-trip
- Rewritten `agent_reminder_emission_test.go`: per-hook dispatch, emission tagged with `hook`+`name`, the orphan-tool-call guard, and the provider interface via the generated fake
- Headless `TestDispatchHooks_Reminders` in `cmd/agent_test.go`
- Config regenerated via `infer init --overwrite`; user-curated `.infer/*.yaml` restored

Not run: a live `infer chat --verbose` smoke test (no gateway/LLM in CI), but the above covers every trigger, hook, the fired-set, the guard, and the emission format.

## Follow-up

A `[DOCS]` ticket for `inference-gateway/docs` is warranted for the new `reminders.yaml` surface and the hook-point catalog.


cmd/agent_test.go
```diff
@@ -1273,175 +1271,61 @@ func TestInjectSystemReminderIfDue(t *testing.T) {
...
-		s.injectSystemReminderIfDue(0)
-
-		if len(s.conversation) != 0 {
-			t.Errorf("expected no messages, got %d", len(s.conversation))
-		}
-		if buf.Len() != 0 {
-			t.Errorf("expected no stream event, got %q", buf.String())
+	t.Run("turns_before_max reminder fires near max at post_session", func(t *testing.T) {
+		s := newSession(true, config.ReminderConfig{Name: "wrap", Text: "wrap up now", Hook: domain.HookPostSession, Trigger: config.ReminderTriggerTurnsBeforeMax, Threshold: 2})
+		s.dispatchHooks(domain.HookPostSession, 9) // maxTurns 10, (10-9)=1 <= threshold 2
```
```suggestion
		s.dispatchHooks(domain.HookPostSession, 9)

File: config/prompts.go [hunk 2]
```diff
@@ -44,18 +45,8 @@ func mergePromptDefaults(loaded, defaults *PromptsConfig) {
...
-		loaded.Agent.SystemReminders.Interval = defaults.Agent.SystemReminders.Interval
-	}
-	if loaded.Agent.SystemReminders.WrapUpText == "" {
-		loaded.Agent.SystemReminders.WrapUpText = defaults.Agent.SystemReminders.WrapUpText
-	}
-	if loaded.Agent.SystemReminders.WrapUpThreshold == 0 {
-		loaded.Agent.SystemReminders.WrapUpThreshold = defaults.Agent.SystemReminders.WrapUpThreshold
-	}
+	// SystemReminders.Reminders is intentionally not backfilled: an explicit user
+	// list wins, and an absent prompts.yaml already yields the defaults via LoadYAML.
```
edenreich:
```suggestion
```

PR #698, opened 2026-06-30
fix(autocomplete): surface arguments for one-of/all-optional tool schemas

edenreich:
## Summary

Fixes #690.

The `!!` direct-tool autocomplete only filled the skeleton with top-level `required` properties, so tools with no top-level `required` array (like the `Agent` tool - a one-of between `tasks` and `description`) showed a bare `Agent()` with no indication of what to pass. The same gap affected any tool whose useful arguments are not top-level `required` (one-of or all-optional schemas).

## Root cause

`generateToolTemplate` (`internal/ui/autocomplete/autocomplete.go`) filled the parentheses with **only** the tool's top-level `required` properties via `extractRequiredArguments`. The `Agent` tool's schema (`internal/agent/tools/agent.go`) has no top-level `required` array - its only `required` is nested under `tasks.items` - so `extractRequiredArguments` returned nothing and the template was `!!Agent()`. The suggestion's `Description` was the generic "Execute <tool> tool directly" and its `Usage` was empty, so no parameter names/types/descriptions were surfaced anywhere.

## Changes

- **`generateToolArguments`** (new, extracted from `generateToolTemplate` to keep nesting under the `nestif` limit) - prefers top-level `required` properties, then **falls back to all top-level properties** so one-of / all-optional schemas still surface their meaningful arguments.
- **`extractAllProperties`** (new) - builds argument templates for every top-level property, `slices.Sort`-ed by name for deterministic output. Returns nil when there are no properties (so parameter-less tools like `WebSearch`/`Tree` keep `!!WebSearch(` unchanged).
- **`generateArgumentTemplate`** - added `array` -> `=[]` and `object` -> `={}` cases (previously both fell through to `=""`).
- **`generateParameterHints`** (new) - builds a compact `name (type, required|optional)` list from the schema's top-level properties and `required` array, sorted by name. Used to enrich the dropdown description.
- **`loadTools`** - appends ` - args: <hints>` to the dropdown `Description` when the tool has parameters.

## Before / after

**Agent tool** (`!! Age`):

| | Before | After |
| --- | --- | --- |
| Skeleton | `Agent()` | `Agent(description="", system_prompt="", tasks=[], type="")` |
| Description | Execute Agent tool directly | Execute Agent tool directly - args: description (string, optional), system_prompt (string, optional), tasks (array, optional), type (string, optional) |

Tools with top-level `required` (`Read`/`Write`/`Bash`) and parameter-less tools (`WebSearch`/`Tree`) are unchanged.

## Test coverage

`TestAutocomplete_ToolsAllOptionalSchema` registers an Agent-shaped schema (one-of, all-optional top-level properties, with an array property and a nested `required` under `tasks.items`) and asserts:
- The skeleton includes all top-level properties with type-appropriate defaults (`=[]` for array, `=""` for string), in sorted order.
- The dropdown description lists each parameter with its type and `optional` qualifier.
- Filtering on `!!Age` and the empty `!!` prefix both work.

## Verification

- [x] `task test` - all packages pass
- [x] `task lint` - 0 issues
- [x] `task fmt` - clean
- [x] pre-commit hooks passed (fmt, lint, trailing whitespace, EOF)

no docs ticket: internal-only TUI autocomplete fix

Fixed by GLM-5.2

### DIFF
--- a/internal/ui/autocomplete/autocomplete_test.go
+++ b/internal/ui/autocomplete/autocomplete_test.go
@@ -600,9 +600,103 @@ func TestAutocomplete_SkillsMidText(t *testing.T) {
 	})
 }
 
-// TestAutocomplete_TabCompletesWithoutSubmitting locks in that Tab only
-// completes a suggestion (with a trailing space) and never auto-submits, even
-// for a no-arg shortcut. The user submits with a deliberate Enter afterwards.
+// TestAutocomplete_ToolsAllOptionalSchema covers the regression in issue #690:
+// a tool whose useful arguments are not top-level "required" (a one-of /
+// all-optional schema like the Agent tool) must still surface its arguments in
+// the !! skeleton and the dropdown description, rather than showing a bare
+// "Tool()" with a generic "Execute <tool> tool directly" line.
+func TestAutocomplete_ToolsAllOptionalSchema(t *testing.T) {
+	mockRegistry := &uimocks.FakeShortcutRegistry{}
+	mockRegistry.GetAllReturns([]shortcuts.Shortcut{})
+
+	mockToolService := &domainmocks.FakeToolService{}
+
+	agentDesc := "Spawn local subagents in parallel"
+
+	// Agent-shaped schema: top-level properties with NO top-level "required"
+	// array (a one-of between `tasks` and `description`). The only "required"
+	// is nested under tasks.items, which the !! template builder must not rely
+	// on. Includes an array-typed property to exercise the new array handling.
+	agentParams := sdk.FunctionParameters(map[string]any{
+		"type": "object",
+		"properties": map[string]any{
+			"tasks": map[string]any{
+				"type":        "array",
+				"description": "Subagent tasks to run in parallel",
+				"items": map[string]any{
+					"type": "object",
+					"properties": map[string]any{
+						"description": map[string]any{"type": "string"},
+					},
+					"required": []string{"description"},
+				},
+			},
+			"description": map[string]any{
+				"type":        "string",
+				"description": "Shorthand for a single subagent task",
+			},
+			"system_prompt": map[string]any{
+				"type":        "string",
+				"description": "Optional system prompt for the single-task form",
+			},
+			"type": map[string]any{
+				"type":        "string",
+				"enum":        []string{"ReadOnly", "ReadWrite"},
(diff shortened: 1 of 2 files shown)


internal/ui/autocomplete/autocomplete_test.go
```diff
@@ -600,9 +600,103 @@ func TestAutocomplete_SkillsMidText(t *testing.T) {
...
+	mockRegistry.GetAllReturns([]shortcuts.Shortcut{})
+
+	mockToolService := &domainmocks.FakeToolService{}
+
+	agentDesc := "Spawn local subagents in parallel"
+
+	// Agent-shaped schema: top-level properties with NO top-level "required"
+	// array (a one-of between `tasks` and `description`). The only "required"
+	// is nested under tasks.items, which the !! template builder must not rely
+	// on. Includes an array-typed property to exercise the new array handling.
```
edenreich:
```suggestion
```

internal/ui/autocomplete/autocomplete_test.go
```diff
@@ -600,9 +600,103 @@ func TestAutocomplete_SkillsMidText(t *testing.T) {
...
+	})
+
+	theme := &uimocks.FakeTheme{}
+	theme.GetDimColorReturns("#808080")
+	theme.GetAccentColorReturns("#FF00FF")
+
+	ac := autocomplete.NewAutocomplete(theme, mockRegistry)
+	ac.SetToolService(mockToolService)
+	// Use a wide terminal so the description column is wide enough for the full
+	// parameter hint list (it is truncated to descWidth during rendering).
```
edenreich:
```suggestion
```

internal/ui/autocomplete/autocomplete_test.go
```diff
@@ -600,9 +600,103 @@ func TestAutocomplete_SkillsMidText(t *testing.T) {
...
+	ac.SetWidth(240)
+
+	t.Run("skeleton includes all top-level properties with type-appropriate defaults", func(t *testing.T) {
+		ac.Update("!!Age", 5)
+		assert.True(t, ac.IsVisible(), "!!Age should match the Agent tool")
+		// Properties are surfaced in sorted order; array -> [], string -> "".
```
edenreich:
```suggestion
```

PR #700, opened 2026-06-30
refactor(tui): single event-loop ingress - one notifier, no self-rescheduling pollers

edenreich:
## Summary

Replaces every self-rescheduling loop that fed the single Bubble Tea `Update` loop with **one push ingress**: a tea-free `domain.UINotifier` backed by `(*tea.Program).Send`, injected into every background producer via a thread-safe container holder. The `Update` loop is now **purely reactive** — it renders on real events, not on a clock — and because there is exactly one ingress, it's instrumented in exactly one place.

Closes #688. `no docs ticket: internal-only refactor`.

## What was removed (all re-arming pollers)

| Ingress | Was | Now |
| --- | --- | --- |
| Queue-drain | forever `tea.Tick(500ms)` → `DrainQueueTickEvent`, self-re-arming | `DrainQueueEvent` pushed once per real trigger; gate-only handler |
| Agent-status | `time.Sleep(500ms)` Cmd re-arm | pushed from the container status callback on change |
| MCP status | `StartMonitoring` returned a channel the app re-armed (`waitForMCPStatusUpdate`/`mcpStatusUpdateWithChannel`) | monitor pushes `MCPServerStatusUpdateEvent` through the notifier |
| File explorer / diff viewer | `tea.Tick(1s)` git/fs re-poll | refresh off in-loop `ToolExecutionCompletedEvent`/`BashCommandCompletedEvent` + view-entry + manual `ctrl+r` |
| Computer-use forwarder | bespoke `program.Send` ×2 | routed through the one notifier |

The only `tea.Tick`s left are the three **one-shot** UI deferrals the design keeps (todo auto-collapse, two background-row removals).

## Design

- `internal/domain/ui_notifier.go` — `UINotifier interface{ Notify(any) }` + `NoopUINotifier` (useful zero value) + `NotifierFunc` (test adapter). `any`, not `tea.Msg`, so the domain stays tea-free.
- `internal/container` — a thread-safe `uiNotifierHolder` (RWMutex + swappable inner, default Noop) is the single synchronized primitive. Producers capture it at construction and never lock; `SetUINotifier` swaps the inner once at startup.
- `cmd/chat.go` — `programNotifier{program}` is the **only** place `(*tea.Program).Send` is ever called; `services.SetUINotifier(...)` runs after `tea.NewProgram` and before `program.Run`.

## Correctness notes

- **No-strand guarantee.** `HandleDrainQueueEvent` is a pure gate (returns `nil` in every no-op branch — regression guard against re-introducing a ticker). Drains are pushed from three sources: supervisor `enqueue`, terminal turn completion (`HandleChatCompleteEvent`), and a single central chat-re-entry edge in `Update`. Added the symmetric `EndToolExecution()` on terminal completion so the agent is reliably idle when the drain runs — nothing strands without a poller.
- **No deadlock.** `(*tea.Program).Send` is unbuffered and blocks until `Run` consumes, so `Notify` is only ever called from producer goroutines; the MCP initial-status emit now runs in a goroutine.
- **Interface change.** `MCPManager.StartMonitoring` drops its channel return → `mcp_manager_test.go` rewritten to assert notifier pushes via a recording notifier (MCPManager has no generated counterfeiter fake).

## Observability

One slow-update warn around the `Update` dispatch (`SlowUpdateThreshold`) and a one-time long-running-job warn in the supervisor cleanup sweep (`JobRunningLongThreshold`, reusing `meta.StartedAt`).

## Acceptance

- [x] `program.Send` called in exactly one place (grep-proven).
- [x] Zero self-rescheduling pollers / re-arming wait-Cmds remain; only one-shot deferrals survive.
- [x] `HandleDrainQueueEvent` is gate-only; asserted in `chat_handler_drain_test.go`.
- [x] Agent-status, MCP-status, `/tasks` update on real change pushed through the notifier — no clock.
- [x] Explorer/diff refresh off in-loop tool/bash events + view-entry + manual key; 1s `tickCmd`s deleted; **no fsnotify in the TUI**.
- [x] New tests: supervisor fires on enqueue and on status change; terminal-emit decision; readiness/MCP push; `EndToolExecution` symmetry; slow-update + long-job warns via an observed zap core.
- [x] `task build`, `task test`, `task lint` (0 issues), `task fmt` green; `go test -race` clean on touched packages.

## Verification left for review

The interactive runtime checks from the issue (idle on `/model` shows no periodic re-render, drain-while-modal scenarios, live `/tasks`, diff/explorer refresh off agent edits) need a running gateway + TTY and are best confirmed manually; the behavior is covered by the unit tests above.

## Out of scope / unchanged

One-shot UI deferrals (todo auto-collapse, background-row removal), the agent state-machine `CheckingQueue` in-turn drain, and `EventBridge`'s floating-window multicast.

### DIFF
--- a/internal/services/jobs/supervisor.go
+++ b/internal/services/jobs/supervisor.go
@@ -71,12 +71,10 @@ func NewSupervisor(messageQueue domain.MessageQueue, conversationRepo domain.Con
 	}
 }
 
-// notify pushes an event to the UI loop. It is nil-safe so a zero-value
-// Supervisor (e.g. one built directly in a test) never panics.
+// notify pushes an event to the UI loop. NewSupervisor defaults a nil notifier to
+// NoopUINotifier, so s.notifier is always non-nil here.
 func (s *Supervisor) notify(event any) {
-	if s.notifier != nil {
-		s.notifier.Notify(event)
-	}
+	s.notifier.Notify(event)
 }
 
 // SetConversationRepo wires the conversation repository used to format finished
--- a/internal/services/jobs/supervisor.go
+++ b/internal/services/jobs/supervisor.go
@@ -18,6 +18,7 @@ import (
 
 	sdk "github.com/inference-gateway/sdk"
 
+	constants "github.com/inference-gateway/cli/internal/constants"
 	domain "github.com/inference-gateway/cli/internal/domain"
 	logger "github.com/inference-gateway/cli/internal/logger"
 )
@@ -26,6 +27,7 @@ import (
 type Supervisor struct {
 	messageQueue     domain.MessageQueue
 	conversationRepo domain.ConversationRepository
+	notifier         domain.UINotifier
 
 	mu              sync.RWMutex
 	jobs            map[string]*supervised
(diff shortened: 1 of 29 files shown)


internal/services/jobs/supervisor.go
```diff
@@ -46,21 +48,37 @@ type supervised struct {
 	completedAt *time.Time
 	lastNote    string
 	cancel      context.CancelFunc
+	warnedLong  bool // a "job running long" warning has been emitted once
```
edenreich:
```suggestion
	warnedLong  bool
```

PR #724, opened 2026-07-02
fix(keybindings): stop "unknown keybinding action" log flood

edenreich:
## Summary

Fixes the flood of `unknown keybinding action` warnings in `~/.infer/logs/app-<date>.log` - about **84 lines per `infer chat` launch** (42 warn + 42 error), even with a stock `infer init` config.

## Root cause

- **Structural drift (the flood).** `config.GetDefaultKeybindings()` includes 41 namespace-path actions (`chat_focus_attachments`, all `diff_viewer_*`, all `explorer_*`) that are consumed directly via `config.ResolveNamespaceBindings` and are never registered in the runtime `Registry`. `(*Registry).ApplyConfigOverrides` warned for every config ID absent from the registry, so it false-flagged all 41 on every launch. The CLI validator `getValidActionIDs()` shared the same blind spot.
- **Logger misuse (doubled the noise).** `registry.go:364` called `logger.Warn("... '%s' ...", actionID)`, but `logger.Warn(msg, args...)` forwards to zap's key/value sugared API (`Warnw`). So `%s` never expanded and the trailing `actionID` became a dangling key - which is why zap added a second, `error`-level `Ignored key without a value.` line.

## Fix

- New `config.DefaultKeybindingActionIDs()` - the config-side valid-ID set.
- `ApplyConfigOverrides` now warns only for IDs in **neither** the registry **nor** the defaults (exact-ID match, so genuine typos and stale/removed IDs still warn once), and logs structured (`"action", id`). The sibling misuse at `registry.go:36` is fixed too.
- New `Registry.KnownActionIDs()` (registry union defaults); the CLI validator uses it so `infer keybindings validate/set/enable/disable` agree with the runtime.
- Regression tests: a stock config is silent; a genuine/namespace typo warns exactly once with a structured `action` field and zero `Ignored key` lines; the validator includes namespace-path IDs.

## Verification

- `task build`, `task lint` (0 issues), `go test ./...` all pass.
- `infer keybindings validate` against a real stale config: **42 -> 1** unknown ID (only the genuinely-removed `plan_approval_..._accept_and_auto_approve` from #714). Runtime flood drops from 84 lines/launch to at most 1 clean structured warning.

## Notes

- No config regeneration - `GetDefaultKeybindings()` output is unchanged.
- The doubled `plan_approval_plan_approval_` prefix is intentionally left as-is (identical between registry and defaults; renaming would break existing user configs).
- The same printf-style structured-logger misuse exists at ~18 other call sites (same defect class, malformed logs); tracked separately in #723.

Closes #722

### DIFF
--- a/internal/ui/keybinding/registry_warnings_test.go
+++ b/internal/ui/keybinding/registry_warnings_test.go
@@ -29,7 +29,7 @@ func TestApplyConfigOverrides_UnknownActionWarnings(t *testing.T) {
 		name      string
 		bindings  map[string]config.KeyBindingEntry
 		wantWarns int
-		wantField string // when set, asserted as the warn's action field
+		wantField string
 	}{
 		{
 			name:      "fresh default config is silent",
@@ -59,8 +59,6 @@ func TestApplyConfigOverrides_UnknownActionWarnings(t *testing.T) {
 			wantField: "diff_viewer_bogus",
 		},
 		{
-			// #714 renamed this action; a stale config keeps the old (doubled-prefix)
-			// ID, which is genuinely unknown now and should warn exactly once.
 			name:      "stale renamed action warns once",
 			bindings:  map[string]config.KeyBindingEntry{"plan_approval_plan_approval_accept_and_auto_approve": {Keys: []string{"a"}, Enabled: &enabled}},
 			wantWarns: 1,
@@ -78,15 +76,13 @@ func TestApplyConfigOverrides_UnknownActionWarnings(t *testing.T) {
 			cfg := &config.Config{}
 			cfg.Chat.Keybindings = config.KeybindingsConfig{Enabled: true, Bindings: tt.bindings}
 
-			_ = keybinding.NewRegistry(cfg) // triggers ApplyConfigOverrides
+			_ = keybinding.NewRegistry(cfg)
 
 			unknown := logs.FilterMessage("unknown keybinding action in config, ignoring").All()
 			if len(unknown) != tt.wantWarns {
 				t.Fatalf("unknown-keybinding warns = %d, want %d (all logs: %v)", len(unknown), tt.wantWarns, logs.All())
 			}
 
-			// Bug A regression guard: the structured logger must not emit zap's
-			// "Ignored key without a value." error from a dangling printf arg.
 			if n := logs.FilterMessageSnippet("Ignored key").Len(); n != 0 {
 				t.Fatalf("got %d spurious 'Ignored key without a value.' lines (Bug A regression)", n)
 			}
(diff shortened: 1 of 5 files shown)


internal/ui/keybinding/registry_warnings_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+			_ = keybinding.NewRegistry(cfg) // triggers ApplyConfigOverrides
+
+			unknown := logs.FilterMessage("unknown keybinding action in config, ignoring").All()
+			if len(unknown) != tt.wantWarns {
+				t.Fatalf("unknown-keybinding warns = %d, want %d (all logs: %v)", len(unknown), tt.wantWarns, logs.All())
+			}
+
+			// Bug A regression guard: the structured logger must not emit zap's
+			// "Ignored key without a value." error from a dangling printf arg.
```
edenreich:
```suggestion
```

internal/ui/keybinding/registry_warnings_test.go
```diff
@@ -0,0 +1,101 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			core, logs := observer.New(zapcore.WarnLevel)
+			prev := logger.GetGlobalLogger()
+			logger.SetGlobalLogger(zap.New(core))
+			defer logger.SetGlobalLogger(prev)
+
+			cfg := &config.Config{}
+			cfg.Chat.Keybindings = config.KeybindingsConfig{Enabled: true, Bindings: tt.bindings}
+
+			_ = keybinding.NewRegistry(cfg) // triggers ApplyConfigOverrides
```
edenreich:
```suggestion
			_ = keybinding.NewRegistry(cfg)
```

internal/ui/keybinding/registry_warnings_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		{
+			name:      "namespace typo still warns once",
+			bindings:  map[string]config.KeyBindingEntry{"diff_viewer_bogus": {Keys: []string{"ctrl+x"}, Enabled: &enabled}},
+			wantWarns: 1,
+			wantField: "diff_viewer_bogus",
+		},
+		{
+			// #714 renamed this action; a stale config keeps the old (doubled-prefix)
+			// ID, which is genuinely unknown now and should warn exactly once.
```
edenreich:
```suggestion
```

internal/ui/keybinding/registry_warnings_test.go
```diff
@@ -0,0 +1,101 @@
...
+//
+// The subtests mutate the package-global sugared logger, so they must not run in
+// parallel.
+func TestApplyConfigOverrides_UnknownActionWarnings(t *testing.T) {
+	enabled := true
+	tests := []struct {
+		name      string
+		bindings  map[string]config.KeyBindingEntry
+		wantWarns int
+		wantField string // when set, asserted as the warn's action field
```
edenreich:
```suggestion
		wantField string
```

PR #729, opened 2026-07-02
fix(logging): use structured key/value form at misused call sites

edenreich:
## Summary

`internal/logger`'s `Debug/Info/Warn/Error/Fatal(msg string, args ...any)` forward to zap's sugared
**key/value** API (`Warnw`/`Errorw`/…). ~18 call sites instead passed a **printf-style** format string
(`%s`/`%v`/`%d`) with trailing positional args, so:

- the verbs were never expanded (a literal `%s` landed in the log), and
- the trailing args were misread as key/value pairs — an **odd** count left a dangling key, making zap
  emit an extra `error`-level `Ignored key without a value.` line.

This is the same defect class as the keybinding log flood (#722, fixed in #724, which corrected the two
worst offenders and tracked the rest in #723). This PR closes the remaining sites.

## Changes

- **Convert 18 call sites to structured key/value form**, following the #724 precedent
  (`logger.Warn("failed to …", "error", err)`). Multi-word keys use the repo's snake_case house style
  (`chat_id`, `original_bytes`).
  - `internal/ui/keybinding/actions.go` (7), `internal/agent/agent_utils.go` (5),
    `internal/services/channels/telegram.go` (4), `internal/app/chat.go` (1),
    `internal/container/container.go` (1)
- **Enable the `loggercheck` linter** (`no-printf-like` + wrapper rules for `internal/logger.*`) so
  this class cannot regress.

The 9 other logger calls that contain a `%` were verified as correct `fmt.Sprintf` usage (message
pre-formatted, or building a properly-keyed value) and left untouched.

## Verification

- `task build`, `task test` (full suite), and `task lint` (**0 issues**, loggercheck active) all pass.
- Confirmed the guard is live: a deliberate misuse is flagged by loggercheck
  (`odd number of arguments passed as key-value pairs for logging`); the fixed tree lints clean.

no docs ticket: internal-only fix

Closes #723

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,8 +37,6 @@ linters:
     nestif:
       min-complexity: 4
     loggercheck:
-      # Guard against printf-style format strings (%s/%v) passed to the
-      # structured key/value logger. Teach loggercheck our wrapper funcs.
       no-printf-like: true
       rules:
         - github.com/inference-gateway/cli/internal/logger.Debug
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,6 +16,7 @@ linters:
     - funlen
     - gocognit
     - nestif
+    - loggercheck
   disable:
     - exhaustruct
     - varnamelen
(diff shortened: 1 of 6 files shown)


.golangci.yml
```diff
@@ -35,6 +36,16 @@ linters:
       min-complexity: 45
     nestif:
       min-complexity: 4
+    loggercheck:
+      # Guard against printf-style format strings (%s/%v) passed to the
+      # structured key/value logger. Teach loggercheck our wrapper funcs.
```
edenreich:
```suggestion
```

PR #767, opened 2026-07-06
fix(debug): sync memory in before rendering the agent system prompt

edenreich:
## Summary

`infer debug agent system_prompt` renders the prompt straight from disk without syncing memory in, so on a fresh machine with the git memory backend configured (e.g. a CI runner probing the prompt before the agent starts) the `PERSISTENT MEMORY INDEX` section is silently omitted — while the real agent, which runs `SyncIn` in its pre-session hook before building the prompt, does receive it. The debug render under-reports the agent's reality.

Observed via infer-action, which probes the merged prompt before spawning `infer agent`: [docs run 28809013238](https://github.com/inference-gateway/docs/actions/runs/28809013238/job/85431796214) shows the full context block but no memory index despite the git backend being fully configured.

## Change

The debug command now calls `SyncIn` on the container's memory backend before `BuildSystemPrompt()`, mirroring the headless agent's pre-session hook. Fail-soft like the agent (the backend warns and continues); the local backend is a no-op, and the git backend self-gates on `sync.on_start` and is `sync.Once`-idempotent — so behavior is unchanged except when a configured git memory repo hasn't been pulled yet.

## Test

`cmd/debug_test.go` reproduces the fresh-runner case: a seeded local bare repo as the memory remote, a non-existent memory dir, and asserts the rendered prompt contains `PERSISTENT MEMORY INDEX` and the seeded fact.

### DIFF
--- a/cmd/debug_test.go
+++ b/cmd/debug_test.go
@@ -7,8 +7,9 @@ import (
 	"path/filepath"
 	"testing"
 
-	config "github.com/inference-gateway/cli/config"
 	require "github.com/stretchr/testify/require"
+
+	config "github.com/inference-gateway/cli/config"
 )
 
 // The fresh-CI-runner case: the git memory backend is configured but nothing
--- a/cmd/debug_test.go
+++ b/cmd/debug_test.go
@@ -15,7 +15,6 @@ import (
 // has cloned the memory repo yet. renderAgentSystemPrompt must SyncIn before
 // building the prompt so the PERSISTENT MEMORY INDEX section renders.
 func TestRenderAgentSystemPrompt_SyncsGitMemoryIn(t *testing.T) {
-	// Isolate git from the developer's config (identity, signing).
 	t.Setenv("GIT_CONFIG_GLOBAL", filepath.Join(t.TempDir(), "no-such-gitconfig"))
 	t.Setenv("GIT_CONFIG_SYSTEM", os.DevNull)
 	t.Setenv("GIT_AUTHOR_NAME", "infer-test")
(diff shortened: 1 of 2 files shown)


cmd/debug_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	config "github.com/inference-gateway/cli/config"
+	require "github.com/stretchr/testify/require"
+)
+
+// The fresh-CI-runner case: the git memory backend is configured but nothing
+// has cloned the memory repo yet. renderAgentSystemPrompt must SyncIn before
+// building the prompt so the PERSISTENT MEMORY INDEX section renders.
+func TestRenderAgentSystemPrompt_SyncsGitMemoryIn(t *testing.T) {
+	// Isolate git from the developer's config (identity, signing).
```
edenreich:
```suggestion
```

cmd/debug_test.go
```diff
@@ -0,0 +1,67 @@
...
+		cmd := exec.Command("git", args...)
+		if dir != "" {
+			cmd.Dir = dir
+		}
+		out, err := cmd.CombinedOutput()
+		require.NoError(t, err, "git %v: %s", args, out)
+	}
+
+	// Bare remote seeded with a MEMORY.md global fact (no project prefix, so
+	// filterMemoryIndex keeps it regardless of the detected project slug).
```
edenreich:
```suggestion
```

cmd/debug_test.go
```diff
@@ -0,0 +1,67 @@
...
+	work := t.TempDir()
+	mustGit("", "clone", bare, work)
+	mustGit(work, "checkout", "-B", "main")
+	require.NoError(t, os.WriteFile(filepath.Join(work, "MEMORY.md"),
+		[]byte("# Memory Index\n\n- [ci-fact](ci-fact.md) - proves debug sync-in ran\n"), 0o600))
+	mustGit(work, "add", "-A")
+	mustGit(work, "commit", "-m", "seed memory")
+	mustGit(work, "push", "-u", "origin", "main")
+
+	// Memory dir does NOT exist yet.
```
edenreich:
```suggestion
```

cmd/debug_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	config "github.com/inference-gateway/cli/config"
+	require "github.com/stretchr/testify/require"
```
edenreich, line 10:
```suggestion

	require "github.com/stretchr/testify/require"

	config "github.com/inference-gateway/cli/config"
```

PR #860, opened 2026-07-11
refactor(agent): unify chat and headless token accumulation

edenreich:
Closes #835

## What

Routes the sync `AgentService.Run` path (headless) through the same `storeIterationMetrics` accumulator the streaming (chat) path already uses, and deletes the headless duplicates. The shared sink was already there — the conversation repository's `AddTokenUsage`/`GetSessionTokens`/`GetSessionCostStats` (what `/cost` reads); the root cause was that `Run` skipped it.

- `internal/agent/agent.go`: `Run` now calls `storeIterationMetrics` (tokenizer polyfill included) and returns the effective usage in `ChatSyncResponse.Usage`, so per-message output matches what was accumulated.
- `cmd/agent.go`: drops the four inline counters, the ad-hoc `AddTokenUsage` goroutine, and the direct `PricingService` cost recomputation; `emitSessionStats` reads the repo's session token and cost stats. `--no-save` now calls `SetAutoSave(false)` so accumulation stays in-memory only.
- `internal/services/persistent_conversation.go`: `SetAutoSave` takes the metadata mutex (one-line hardening).
- New integration test `TestSyncAndStreamAccumulateIdenticalSessionTokens`: both paths against the mock gateway's `usage-report` scenario accumulate identical session totals (142 tokens).

## Acceptance criteria (#835)

- ✅ Single accumulation code path; headless totals equal chat totals for identical mock scenarios (asserted via `usage-report`, 142 tokens)
- ✅ `/cost` and the headless final stats read from the same source (the conversation repo session stats)

## Intended behavior changes

- `session_stats` is now **session/conversation-scoped** (same numbers `/cost` shows) instead of run-scoped. Fresh sessions (CI, subagents, scheduler/heartbeat) are unchanged; resumed sessions (`--session-id` under channels) now include restored history; a mid-run rollover resets totals with the conversation. No programmatic consumer parses these values (agentrunner ignores `session_stats` lines).
- A usage-less provider response previously didn't count toward headless `requests`; the polyfill now counts it with estimated tokens — chat parity, the point of the change.
- `--no-save` + resumed session + mid-run rollover: `StartNewConversation` saves explicitly (ignores auto-save), and the persisted metadata's token stats now include the `--no-save` run's usage. Pre-existing save path, marginally widened.

## Verification

- `task test`, `task test:e2e` (race), `task lint` all green.
- Manual: `INFER_GATEWAY_MOCK=true infer agent "report your usage" --no-save` emits `session_stats` with 142+15 tokens and repo-computed cost; storage untouched under `--no-save`; save-enabled run persists the same accumulated stats.

no docs ticket: internal-only refactor

### DIFF
--- a/cmd/agent_test.go
+++ b/cmd/agent_test.go
@@ -684,8 +684,6 @@ func TestEmitSessionStatsFullLine(t *testing.T) {
 	session := &AgentSession{
 		model:  "deepseek/deepseek-v4-flash",
 		config: &config.Config{Pricing: config.PricingConfig{Currency: "USD"}},
-		// 7 requests at 3000/180/3180 tokens, each costing 0.25/0.125/0.375
-		// (binary-exact so the accumulated sums compare with ==).
 		conversationRepo: seededRepo(t, fakePricing(0.25, 0.125, 0.375),
 			"deepseek/deepseek-v4-flash", 7, 3000, 180, 3180),
 	}
--- a/cmd/agent_test.go
+++ b/cmd/agent_test.go
@@ -666,15 +666,28 @@ func decodeStatsLine(t *testing.T, session *AgentSession) sessionStatsLine {
 	return got
 }
 
+// seededRepo returns an in-memory conversation repository - the shared sink the
+// agent service accumulates into - seeded with n AddTokenUsage calls of the
+// given per-request token counts.
+func seededRepo(t *testing.T, pricing domain.PricingService, model string, n, prompt, completion, total int) *services.InMemoryConversationRepository {
+	t.Helper()
+	repo := services.NewInMemoryConversationRepository(nil, pricing)
+	for range n {
+		if err := repo.AddTokenUsage(model, prompt, completion, total); err != nil {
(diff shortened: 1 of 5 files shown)


cmd/agent_test.go
```diff
@@ -666,15 +666,28 @@ func decodeStatsLine(t *testing.T, session *AgentSession) sessionStatsLine {
...
-		pricingService:        fakePricing(0.0021, 0.0008, 0.0029),
-		config:                &config.Config{Pricing: config.PricingConfig{Currency: "USD"}},
-		totalPromptTokens:     21000,
-		totalCompletionTokens: 1260,
-		totalTokens:           22260,
-		requestCount:          7,
+		model:  "deepseek/deepseek-v4-flash",
+		config: &config.Config{Pricing: config.PricingConfig{Currency: "USD"}},
+		// 7 requests at 3000/180/3180 tokens, each costing 0.25/0.125/0.375
+		// (binary-exact so the accumulated sums compare with ==).
```
```suggestion
```

PR #885, opened 2026-07-12
fix(tui): keep arg highlight on truncation and expand summary with ctrl+o

edenreich:
## Summary

Two fixes to the approval-box tool-call summary (follow-up to #882).

**1. Highlight survives truncation.** The arg highlight split the summary at the first `(` and last `)`, but when a long command was truncated the closing `)` was cut off, so `LastIndexByte` returned -1 and the whole line fell back to all-dim. Now a missing `)` highlights the arguments through to the end of the truncated string — so the highlight is consistent instead of "sometimes works, sometimes not".

**2. `ctrl+o` expands the args.** For non-diff tools the summary was always a truncated one-liner regardless of the expanded state. `ctrl+o` (already wired to the approval box for diff previews) now soft-wraps the full argument list so a long command can be reviewed in full before approving. Collapsed shows a `(ctrl+o to expand)` hint when truncated; expanded shows `(ctrl+o to collapse)`.

## Testing

`go build ./...`, `go test ./internal/ui/components/`, pre-commit hooks all pass.

### DIFF
--- a/internal/ui/components/approval_box_view.go
+++ b/internal/ui/components/approval_box_view.go
@@ -225,13 +225,45 @@ func (av *ApprovalBoxView) renderBody(tc *sdk.ChatCompletionMessageToolCall) str
 // what will be executed. Truncation happens on the plain string first so the ANSI
 // colour codes never throw off the width budget.
 func (av *ApprovalBoxView) renderSummary(tc *sdk.ChatCompletionMessageToolCall) string {
+	full := av.toolCallSummary(tc)
+	budget := av.summaryBudget()
+	oneLine := formatting.TruncateText(full, budget)
+	fits := oneLine == full
+
+	// ctrl+o (ToggleExpanded) soft-wraps the full argument list instead of the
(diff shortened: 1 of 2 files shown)


internal/ui/components/approval_box_view.go
```diff
@@ -225,15 +245,47 @@ func (av *ApprovalBoxView) renderBody(tc *sdk.ChatCompletionMessageToolCall) str
...
+	oneLine := formatting.TruncateText(full, budget)
+	fits := oneLine == full
+
+	// ctrl+o (ToggleExpanded) soft-wraps the full argument list instead of the
+	// truncated one-liner, so a long command can be reviewed in full before
+	// approving — mirroring the diff-preview expansion. wordwrap is ANSI-aware and
+	// the terminal keeps the accent colour active across wrapped lines, so we can
+	// colour first and wrap after.
+	// ponytail: unbounded height when expanded; route through capLines() if a
+	// pathologically long command ever blows out the box.
```
edenreich:
```suggestion
```

PR #887, opened 2026-07-12
refactor(storage): shared SQL core for sqlite and postgres

edenreich:
## Summary

Folds the `sqlite` and `postgres` conversation-storage backends onto one shared SQL core, per #839.

Exploration showed the two backends were **not** dialect twins as the issue assumed: `sqlite` used a single-table, embedded-`messages`-JSON-blob schema (shared verbatim by `d1`), while `postgres` used a normalized two-table schema. `postgres` was also **broken** — `ListConversations` / `ListConversationsNeedingTitles` selected a `summary` column that its schema never created and mismatched the row scan (12 destinations for 13 columns), and it had no happy-path tests, so it shipped undetected.

The fix is to collapse `postgres` onto the same single-table schema `sqlite` and `d1` already share:

- **New `internal/infra/storage/sql_store.go`** — one `sqlStore{db, dialect}` holds every SQL method once. `sqlite.go` and `postgres.go` become thin constructors embedding `*sqlStore`; the only dialect difference left is placeholder style, handled by a small `rebind` (`?` → `$N`). `postgres.go` drops 531 → ~80 lines, `sqlite.go` 545 → ~65.
- **`postgres_migrations.go`** rewritten to mirror the single-table schema (postgres types), dropping `conversation_entries`. Replacing the hand-written postgres save/load/list with the shared core removes the broken list query outright.
- **Shared conformance suite (`conformance_test.go`)** extracted from the sqlite/d1 tests that used to be copy-pasted. It runs against `sqlite` (temp file) and `d1` (httptest mock) unconditionally, and against a real `postgres` gated on `INFER_TEST_POSTGRES_DSN` — the postgres backend's first happy-path coverage. A CI `postgres` service-container job wires that DSN.

Backends decision: all six retained. `d1` already shared the SQL core; `redis`/`memory`/`jsonl` are untouched.

## Verification

- `task test` (full suite) + `go test -race ./internal/infra/storage/...`, `go vet`, `task lint` (0 issues) — all green.
- Against a real PostgreSQL: conformance suite passes, `infer migrate --status` works, and a full `infer agent` → postgres → `infer conversations list` round-trip renders the conversation (the list path that previously failed).
- `sqlite` `migrate --status` regression check passes (the `cmd/migrate.go` concrete-type switch survives the embedding).

Net: −453 LOC across the storage package.

no docs ticket: internal-only refactor (in-repo storage docs updated here)

Closes #839


internal/infra/storage/postgres_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	cfg := parsePostgresDSN(t, dsn)
+
+	runConversationStorageConformance(t, func(t *testing.T) ConversationStorage {
+		storage, err := NewPostgresStorage(cfg)
+		require.NoError(t, err)
+		t.Cleanup(func() { _ = storage.Close() })
+
+		// Isolate each conformance group: the shared server persists rows across
+		// groups (and runs), so start every one from a clean slate.
```
edenreich:
```suggestion
```

PR #890, opened 2026-07-12
feat(skills): ship built-in tmux skill via seed-if-absent

edenreich:
## Summary

Closes #827.

Ships built-in skills embedded in the binary and seeded into
`~/.infer/skills/<name>/` on `infer init` (**seed-if-absent**, so user edits
survive), reusing the existing skills loader end-to-end — no new scope, no new
config. First built-in is a **`tmux`** skill teaching the agent to drive
interactive terminal programs.

## Behavior

- The `tmux` skill directs the agent to **reuse the current tmux session and
  split a pane the user can watch** when `$TMUX` is set, and to start a detached
  session only when not already inside tmux.
- Customization is free via existing knobs: edit-in-place (never re-seeded),
  project-scope shadow, `disabled_skills`, `init --overwrite` reset.
- `ReadSubagentScreen` / `SendSubagentInput` descriptions now point at the skill
  for non-subagent programs (orthogonal to the CLI-managed interactive-subagent
  panes).

## Consolidation

- `AGENTS.md` "Driving the chat TUI via tmux" is trimmed to repo/mock-gateway
  glue + a pointer to the skill (single source of truth). `CLAUDE.md` updated to
  match. New `tests/e2e/tmux_tui_test.go` drives the real chat TUI the same way.

## Tests

- `SeedBuiltins` unit tests: seed-when-missing (validates via the loader),
  not-overwritten, overwrite-resets.
- e2e `TestChatTUIViaTmux` (skips cleanly if `tmux` is not on PATH).
- `task test` + `task lint` green.

## Follow-ups (not in this PR)

- Seeding is init-triggered; a plain `infer init` won't re-seed over an
  already-initialized home, so refreshing a single built-in after an update
  needs a manual file replace or `init --overwrite`. A `Load()`-time seed (the
  issue's original suggestion) would make "delete the dir + run any command"
  reseed cleanly — worth a follow-up.
- CI: the e2e tmux test only runs where `tmux` is installed (flox/CI env would
  need it added + a relock).
- `[DOCS]` ticket for `inference-gateway/docs` on built-in skills.


cmd/init.go
```diff
@@ -204,6 +205,16 @@ func initializeProject(cmd *cobra.Command) error { //nolint:funlen,gocyclo,cyclo
 
 	userspaceOverwrite := overwrite && !project
 
+	// Seed the built-in skills into the user-scope skills dir (~/.infer/skills),
+	// the same directory the loader reads user skills from. Seed-if-absent, so a
+	// user's edits survive a re-init; --overwrite (home mode) resets to shipped
+	// defaults. Skipped in --project mode - built-ins are user-scope, not project.
```
edenreich:
```suggestion
```

PR #927, opened 2026-07-17
feat(channels): rich Telegram messaging - photos, markdown, slash commands, and per-conversation /stats

edenreich:
## Overview

Turns the Telegram channel into a proper remote control for the agent: replies render as rich messages (photos + markdown), a slash-command menu manages conversations, `/stats` is scoped per-conversation with a phone-friendly layout, and WebFetch no longer poisons the LLM context with binary bodies (and can surface a fetched image straight into the chat).

### Rich messages

- Agent replies render Markdown as Telegram HTML — bold, headers, fenced/inline code, and pipe tables column-aligned via `text/tabwriter` so they stay readable on a phone instead of arriving as raw markdown.
- Local image references in a reply (``, ``, or a bare absolute image path) are uploaded as native photos via `SendPhoto`, read off the filesystem the daemon and the agent subprocess share.

### Slash commands & session management

- `/new` starts a fresh conversation and best-effort wipes the recent chat messages; `/conversations` lists past conversations as tappable buttons to switch context (or `/conversations <id-prefix>`); `/help` lists the channel commands plus the TUI-only ones. Unknown `/text` still falls through to the agent as a normal message.
- Requires persistent storage (the default); conversation switching is brokered through the daemon so state survives across the per-message agent subprocesses.

### Per-conversation `/stats`

- Channel `/stats` now reports usage for the sender's current conversation (resets on `/new` or a `/conversations` switch) instead of daemon-wide totals. Telemetry datapoints are stamped with `gen_ai.conversation.id` (constant per file, so no cardinality growth) and `Aggregate` filters on it; the host `infer stats` CLI and the chat TUI stay daemon-wide.
- The default channel view is a phone-friendly key/value list that never folds on narrow screens, with 24h/7d buttons and a Table view toggle (or a `table` argument); the chat TUI keeps the aligned tables.

### WebFetch: keep binary out of the LLM context

- A fetched image/PDF/binary body was cast to a string and inlined into the tool result, so the raw bytes entered the LLM context — one browser-agent screenshot pushed a Telegram run past the model's context window (gateway 400: 1,554,486 tokens requested against a 1,048,565 limit). WebFetch now detects non-text content (Content-Type, with a UTF-8 validity fallback), saves it to disk, and returns only a short marker with the path, type, and size; the saved file is byte-exact.
- Under a channel session an image marker also tells the model to reference the saved path as `image`, which the uploader above turns into a photo — so a fetched screenshot appears in the chat without the bytes ever reaching the LLM.

## Testing

- `task test` (unit + integration) and `task lint` are green.
- Verified end-to-end against the `examples/telegram-channel` Docker stack from a phone: rich replies, the slash-command menu, per-conversation `/stats` (list view plus the Table toggle), fetching a page screenshot and having it show up inline, and confirming binary fetches no longer blow the context window.

## Notes

- `examples/telegram-channel/docker-compose.yml` (and a local `docker-compose.override.yml`) are intentionally kept out of this branch.
- Deliberate follow-ups, not in this PR: vision-on-request (letting the model actually "see" a fetched image is a separate opt-in path, so bytes stay out of the context by default) and a `SendDocument` path for non-image files above the 10 MB photo limit.


internal/services/channels/telegram.go
```diff
@@ -426,20 +462,106 @@ func processCallbackQuery(cq *models.CallbackQuery) *domain.InboundMessage {
...
+		add(m[1])
 	}
+	return paths, content
+}
 
+// renderTelegramHTML converts common Markdown (fenced code, inline code, bold,
+// headers) to Telegram HTML. Everything else is escaped so the message never
+// breaks on user content.
+func renderTelegramHTML(md string) string {
+	// ponytail: no italics — *x*/_x_ collide with globs and snake_case in dev output
```
edenreich:
```suggestion
```

PR #963, opened 2026-07-28
feat(chat): add /effort command for runtime anthropic reasoning effort

edenreich:
Stacked on #962 (base branch `feat/942-anthropic-messages-adapter`).

## What

New `/effort` chat command that switches the reasoning effort level at runtime for Anthropic models, delivered on the wire as `output_config.effort` over the native `POST /v1/messages` path. `/effort` with no argument shows the current level and the available ones; `/effort <level>` switches it for subsequent requests. Levels autocomplete (`low | medium | high | xhigh | max`). Non-Anthropic models are rejected with `not supported`.

The CLI hardcodes `low` as its Anthropic default (`config.DefaultAnthropicEffort`) - deliberately below the API default of `high` - so every `/v1/messages` request carries an explicit `output_config.effort`. A status-bar indicator (`Effort: <level>`, own config toggle `chat.status_bar.indicators.effort`) shows the level in effect next to the model, for Anthropic models only.

## How

- `agent.reasoning_effort` config remains the startup seed; the accepted set is extended with `xhigh` and `max` (`config.ReasoningEffortLevels` is now the single source for validation, the setter, and the shortcut).
- `domain.AgentService` gains `SetReasoningEffort`/`GetReasoningEffort`; `AgentServiceImpl` guards the level with an RWMutex and both request paths re-read it per request, so a mid-session switch applies to the next turn.
- The `AnthropicMessages` adapter maps the captured `reasoning_effort` onto `output_config.effort` in `buildMessagesRequest` (`minimal` maps to `low` - Anthropic's scale starts at `low`; unset and unknown values fall back to the hardcoded `low` default).
- The status bar reads the live level through a narrow `effortSource` interface and gates on the current model's provider prefix, so the indicator follows `/model` switches and mid-session `/effort` changes.
- Non-Anthropic providers clamp `xhigh`/`max` to `high` on the chat-completions path as a safety net (relevant when effort was set on an Anthropic model and the user then `/model`-switches away).
- Fixes a builder-chain drop on the sync path: `Run` restarted the chain from the bare client when tools were present (`s.client.WithTools`), silently discarding `MaxTokens`, `ReasoningEffort` and `SkipMCP`; it now chains like the streaming path.
- SDK bumped v1.29.0 -> v1.30.0 (types-only sync with schemas v0.16.0, brings `MessagesOutputConfig`).

## Tests

- Shortcut unit tests: show current/default, lowercased set on Anthropic model, `not supported` rejection on non-Anthropic, invalid-level error, arg bounds.
- Agent unit tests: setter validation/reset, per-provider effort mapping (Anthropic passthrough incl. `minimal`, clamping elsewhere).
- Adapter table test: `output_config.effort` on the marshaled request for every level, `minimal -> low`, absent when unset or unknown.
- Integration (`tests/integration/agent_messages_test.go`): runtime switch to `xhigh` lands on both recorded `/v1/messages` bodies of a tool round-trip.
- `task test`, `task test:e2e`, `task lint`, `task fmt` all green.

## Cross-repo impact

- `schemas` inference-gateway/schemas#161 (merged, v0.16.0) added `output_config.effort` to `CreateMessagesRequest`.
- `sdk` inference-gateway/sdk#148 (merged, v1.30.0) synced the generated types.
- Gateway needs no change: `/v1/messages` forwards bodies verbatim.
- Follow-up: `[DOCS]` issue for inference-gateway/docs covering `/effort` and the extended `agent.reasoning_effort` values.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -757,6 +757,15 @@ func (s *AgentServiceImpl) GetMetrics(requestID string) *domain.ChatMetrics {
 }
 
 // storeIterationMetricsInput holds the data needed for token usage polyfill calculation
+// cacheCreationTokenSource is implemented by clients that report Anthropic
+// cache-creation (cache-write) tokens out of band, since the OpenAI-shaped
+// usage struct has no field for them. The /v1/messages adapter
+// (internal/infra/adapters.AnthropicMessages) is the one implementation;
+// the interface lives here because this package is its consumer.
+type cacheCreationTokenSource interface {
+	TakeCacheCreationTokens() int
+}
+
 type storeIterationMetricsInput struct {
 	inputMessages   []sdk.Message
 	outputContent   string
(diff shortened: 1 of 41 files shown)


internal/agent/agent.go
```diff
@@ -60,6 +61,11 @@ type AgentServiceImpl struct {
 	activeSessions map[string]*sessionCancel
 	sessionMux     sync.RWMutex
 
+	// Runtime reasoning effort, seeded from agent.reasoning_effort and
+	// switchable mid-session via /effort; each request re-reads it.
```
edenreich:
```suggestion
```

PR #983, opened 2026-08-01
fix: deliver A2A completion notifications after every headless turn

edenreich:
## Summary

Fixes the headless A2A completion-notification starvation seen in infer-action run 30717431413: the browser-agent completed its task in 14s and the CLI's poller retrieved `TASK_STATE_COMPLETED` seconds later, but the model never received the notification and burned 15m56s to the job timeout.

**Root cause:** in `infer agent` mode, `drainBackgroundResults` only ran after a turn with **no tool calls**. A model waiting for an A2A result naturally waits by calling tools — in that run a `Wait(120s)`, then a `Wait(300s)` — so the queued completion was never injected: `Wait` times out → model polls/resubmits → more tool calls → starvation. (Chat mode is unaffected; its `CheckingQueue` state drains between turns, which is why local chat testing worked.)

**Fix:** add `BackgroundTasksWaiter.TryDrain()` (non-blocking) and call it after every tool-calling turn in the headless loop, so completions land at the next turn boundary. The blocking wait on quiet turns is unchanged.

Also:
- **`A2A_QUEUE_CLEANUP_INTERVAL=500s`** (overridable via agent env/.env/host env) on agent containers `startContainer` — ADK's 120s default purges completed tasks with no age check, which produced the run's late `A2A_QueryTask` → "task not found" (the task had completed successfully 5 minutes earlier).
- **Stronger `A2A_SubmitTask` result text**: the completion will be injected automatically — do not poll with `A2A_QueryTask` and do not block on the `Wait` tool.
- **AGENTS.md**: `infer chat` / `infer agent` auto-start the gateway and `run: true` agents — never start them manually when testing.

## Verification

- `task test` and `task lint` green; new unit test `TestBackgroundTasksWaiter_TryDrainDoesNotBlock`.
- Headless e2e against the real browser-agent (`go run . agent -m deepseek/deepseek-v4-flash`, full screenshot task): submit once → `✅ A2A Task Completed` notification injected ~14s later → WebFetch downloaded the 901.5 KB artifact to `~/.infer/tmp/` → exit 0.

## Cross-repo impact

- **infer-action**: after release + version-pin bump, CI browser-agent runs should complete instead of hitting `timeout-minutes`.
- **ADK** (follow-ups, not in this PR): age-based completed-task retention instead of the unconditional 120s purge; no OTLP logs exporter exists yet (agent logs can't stream to a collector).

no docs ticket: internal-only fix (restores the behaviour the tool output already promised).

### DIFF
--- a/internal/services/agent_manager.go
+++ b/internal/services/agent_manager.go
@@ -480,10 +480,6 @@ func (am *AgentManager) startContainer(ctx context.Context, agent config.AgentEn
 	gatewayURL := am.determineGatewayURL()
 	env["A2A_AGENT_CLIENT_BASE_URL"] = gatewayURL
 
-	// ADK's default queue cleanup (120s) purges completed tasks with no age
-	// check, so a query a few minutes after completion gets "task not found".
-	// Match the examples' longer interval; agent environment / .env / host env
-	// still override via the resolution below.
 	if _, ok := env["A2A_QUEUE_CLEANUP_INTERVAL"]; !ok {
 		env["A2A_QUEUE_CLEANUP_INTERVAL"] = "500s"
 	}
--- a/internal/services/agent_manager.go
+++ b/internal/services/agent_manager.go
@@ -480,6 +480,14 @@ func (am *AgentManager) startContainer(ctx context.Context, agent config.AgentEn
 	gatewayURL := am.determineGatewayURL()
 	env["A2A_AGENT_CLIENT_BASE_URL"] = gatewayURL
 
+	// ADK's default queue cleanup (120s) purges completed tasks with no age
+	// check, so a query a few minutes after completion gets "task not found".
+	// Match the examples' longer interval; agent environment / .env / host env
+	// still override via the resolution below.
+	if _, ok := env["A2A_QUEUE_CLEANUP_INTERVAL"]; !ok {
+		env["A2A_QUEUE_CLEANUP_INTERVAL"] = "500s"
+	}
+
 	if agent.ArtifactsURL != "" {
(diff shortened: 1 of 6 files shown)


internal/services/agent_manager.go
```diff
@@ -480,6 +480,14 @@ func (am *AgentManager) startContainer(ctx context.Context, agent config.AgentEn
 	gatewayURL := am.determineGatewayURL()
 	env["A2A_AGENT_CLIENT_BASE_URL"] = gatewayURL
 
+	// ADK's default queue cleanup (120s) purges completed tasks with no age
+	// check, so a query a few minutes after completion gets "task not found".
+	// Match the examples' longer interval; agent environment / .env / host env
+	// still override via the resolution below.
```
edenreich:
```suggestion
```

internal/services/background_tasks_waiter_test.go
```diff
@@ -172,3 +172,35 @@ func TestBackgroundTasksWaiter_HasPendingAcrossTaskTypes(t *testing.T) {
...
+	queue.IsEmptyReturnsOnCall(0, false)
+	queue.IsEmptyReturnsOnCall(1, true)
+
+	w := services.NewBackgroundTasksWaiter(minimalA2AConfig(5), "session-1", registry, queue, nil)
+
+	if got := w.TryDrain(); len(got) != 1 {
+		t.Fatalf("expected 1 drained message, got %d", len(got))
+	}
+
+	// Empty queue: returns immediately with nothing, even with tasks pending.
```
edenreich:
```suggestion
```

PR #990, opened 2026-08-02
feat: add named frame sources and vision annotation pipeline

edenreich:
Closes #895

## What

Lets any chat-completions model — including text-only ones like DeepSeek — understand screen and camera frames, via explicit tools instead of base64 image forwarding.

- **Named frame sources**: the screenshot ring buffer becomes the `screen` source; new `directory` sources serve the newest image from a folder (camera frames), with optional retention pruning. Registered on the tool registry, so chat *and* headless `infer agent` get them.
- **Image annotator** (`internal/vlm`): image → scene summary + numbered element list with bounding boxes, via a side-call to a vision model through the gateway (title-generator pattern). The gateway also serves fully local models (Ollama), so offline annotation is just a local provider — no separate engine in the CLI. Default model: `ollama_cloud/qwen3.5-vl`.
- **`GetLatestFrame(source, format)`** replaces `GetLatestScreenshot`: `annotated` returns text *instead of* the image (default when the annotator is enabled), `regular` returns the raw frame. Screen-source annotations include MouseClick-ready element centers.
- **`ImageDecode(image, prompt?)`**: describe any local image file as text, optionally answering a question about it.
- **Attached images carry their saved path**: clipboard pastes and `--files` attachments add a `[image saved at <path>]` text part, so a model without vision can `ImageDecode` the file itself; pasted files are retained (pruned newest-20/24h) instead of deleted on send.

No model fingerprinting: models opt into text by calling the tools, which any orchestrator that speaks chat completions can do.

## Fixes along the way

- Headless `infer agent` silently dropped tool-result images; they now flow and are hoisted into a follow-up user message (tool-role messages stay text-only for Anthropic).
- ComputerUse floating window appended a "running..." line per keep-alive tick; it now updates one line in place with elapsed time and stamps the total on completion.
- The embedded ComputerUse.app is re-extracted on every launch; the previous exists-check kept running a stale copy after CLI upgrades.

## Behavior notes

- `SetScreenshotProvider` (and its MouseClick re-registration vestige) is deleted; MouseClick keeps the shared computer-use rate limiter.
- The annotator parser never fails on malformed model output — it degrades to summary-only; annotation errors degrade to the regular frame with a note, never `Success: false`.

## Tested

- Unit + integration suites green; integration proves the annotated path sends **zero** image parts to the session model while the annotator side-call carries the image.
- Live end-to-end: headless agent with a directory source + Ollama Cloud annotator (accurate shapes/bboxes on a synthetic frame), and interactive chat where `deepseek-v4-flash:cloud` (text-only) described the actual screen — iTerm/tmux panes and code visible in the editor — through the `screen` source + `qwen3.5:cloud` annotator.

## Docs

README ("Frame Sources & Vision Annotation"), docs/tools-reference.md (GetLatestFrame, ImageDecode), docs/configuration-reference.md (Vision Settings, `gateway.vision_enabled` disambiguation). A `[DOCS]` ticket for inference-gateway/docs to follow.

## TODOs

- [ ] Models that supports vision shouldn't need to be instructed to call ImageDecode()
- [ ] Models that supports vision should get the data as base64 included in the content instead of the instructions to use ImageDecode() function
- [ ] Check if the gateway can provide a metadata about the model (if it supports vision or not)
- [ ] Check about MacOS accessibility API or create a follow-up - should reuse existing computer use tools and it should be a parameter passed to allow it to use accessibility API instead of actually moving the mouse or typing with the keyboard, maybe we can define background tasks -> accessibility API it should be pretty good for letting deepseek (non vision model) perform some tasks


cmd/agent.go
```diff
@@ -681,6 +681,11 @@ func (s *AgentSession) buildSDKMessages() []sdk.Message {
 
 	for _, msg := range s.conversation {
 		content := s.buildMessageContent(msg)
+		if msg.Role == "tool" {
+			// Providers (Anthropic) require tool messages to be text-only;
+			// tool-result images are hoisted into a follow-up user message.
```
edenreich:
```suggestion
```

PR #1024, opened 2026-08-05
feat(agent): inject system reminder after 3 identical failing tool calls

edenreich:
## Summary

Weak models can hallucinate a tool call (e.g. a guessed file path) and retry it verbatim forever. In [run 30960749987](https://github.com/inference-gateway/cli/actions/runs/30960749987) `deepseek-v4-flash` retried `Read internal/domain/reminders.go` (a non-existent file) 100+ times — 263 Read calls, 44% failure rate, 42.9M uncached prompt tokens — until the provider returned 400 on context overflow.

## Change

- Track identical failing tool calls (same tool name + exact arguments) in `AgentServiceImpl`, reset per key on success.
- From the 3rd identical failure on, append a `<system-reminder>` to the tool result telling the model to stop retrying and verify its assumptions (list/Grep/Glob for file paths) instead of guessing.
- Applied via a `defer` in `executeToolInternal` — the single choke point for both the standard and event-driven execution paths, so error entries and failed results are all covered, for every tool.

## Testing

- `TestTrackRepeatedFailure` covers the threshold, per-arguments keying, and reset-on-success.
- `task precommit:run` green.

### DIFF
--- a/internal/agent/agent_repeated_failure_test.go
+++ b/internal/agent/agent_repeated_failure_test.go
@@ -4,12 +4,13 @@ import (
 	"strings"
 	"testing"
 
-	domain "github.com/inference-gateway/cli/internal/domain"
 	sdk "github.com/inference-gateway/sdk"
+
+	domain "github.com/inference-gateway/cli/internal/domain"
 )
 
 func TestTrackRepeatedFailure(t *testing.T) {
-	s := &AgentServiceImpl{failedCalls: make(map[string]int)}
+	s := &AgentServiceImpl{}
 	tc := sdk.ChatCompletionMessageToolCall{
 		Function: sdk.ChatCompletionMessageToolCallFunction{
 			Name:      "Read",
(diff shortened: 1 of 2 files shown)


internal/agent/agent_repeated_failure_test.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	if s.trackRepeatedFailure(tc, failed) != "" {
+		t.Fatal("2nd failure should not warn")
+	}
+	note := s.trackRepeatedFailure(tc, failed)
+	if !strings.Contains(note, "<system-reminder>") || !strings.Contains(note, "3 times") {
+		t.Fatalf("3rd failure should warn, got: %q", note)
+	}
+
+	// different args = different key
```
edenreich:
```suggestion
```

internal/agent/agent_repeated_failure_test.go
```diff
@@ -0,0 +1,46 @@
...
+	tc2 := tc
+	tc2.Function.Arguments = `{"file_path":"/other.go"}`
+	if s.trackRepeatedFailure(tc2, failed) != "" {
+		t.Fatal("different args should start a fresh count")
+	}
+
+	// success resets
```
edenreich:
```suggestion
```

PR #1062, opened 2026-08-14
feat(bridge): mirror tool approvals to the browser extension

edenreich:
## What

Bumps the browser-extension bridge protocol to **v2**, adding a tool-approval flow so approvals can be answered from the opentask side panel instead of only the terminal.

Three new frames (documented in `docs/browser-extension-protocol.md`):
- `approval_request` (CLI → extension) — emitted when a tool call needs a decision, carrying `request_id`, `tool_name`, `tool_args`.
- `approval_response` (extension → CLI) — `"approve"` / `"reject"` (unknown actions fail safe to reject).
- `approval_resolved` (CLI → extension) — sent when the request is answered here **or** in the terminal, so the panel clears its card.

## How

`ToolApprovalRequestedEvent` was previously filtered out of the extension chat pump. Now the bridge stashes the pending tool call and forwards it as `approval_request`. An incoming `approval_response` is turned back into the **same** `ToolApprovalResponseEvent` the terminal emits (via `notifier.Notify`), so the existing non-blocking resolve + state-clear naturally wins any terminal/panel race — no new locking, no touching the approval `ResponseChan` directly. Because the agent turn blocks on the approval, the next streamed event reliably signals resolution and triggers `approval_resolved`.

## Companion change

Extension side: inference-gateway/opentask (branch `feat/cli-bridge-sidepanel`).

## Tests

- `TestExtensionBridgeApprovalRoundTrip` — request emitted, response reaches the notifier with the right action + tool call, `approval_resolved` sent.
- `TestExtensionBridgeApprovalUnknownActionRejects` — unknown action fails safe.
- `TestExtensionBridgeApprovalIgnoresUnknownID` — stale/unknown id is a no-op.

`go build ./...`, `go vet`, and the services suite pass.

### DIFF
--- a/internal/services/extension_bridge.go
+++ b/internal/services/extension_bridge.go
@@ -24,8 +24,9 @@ import (
 )
 
 // extensionProtocolVersion is the wire protocol version sent in the hello ack.
-// Documented in docs/browser-extension-protocol.md.
-const extensionProtocolVersion = 1
+// Documented in docs/browser-extension-protocol.md. v2 adds the approval flow
+// (approval_request / approval_response / approval_resolved).
+const extensionProtocolVersion = 2
 
 // Bridge wire messages. One flat envelope per frame, discriminated by Type;
 // unknown types are ignored for forward compatibility.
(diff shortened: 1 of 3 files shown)


internal/services/extension_bridge_test.go
```diff
@@ -7,12 +7,142 @@ import (
...
+}
+
+func TestExtensionBridgeApprovalRoundTrip(t *testing.T) {
+	notifier := &recordingNotifier{}
+	events := macos.NewEventBridge()
+	bridge := startBridge(t, bridgeConfig(), notifier, events)
+	conn := dial(t, bridge)
+	hello(t, conn, "test-token")
+
+	time.Sleep(50 * time.Millisecond) // let the chat pump subscribe
```
edenreich:
```suggestion
	time.Sleep(50 * time.Millisecond)
```

internal/services/extension_bridge.go
```diff
@@ -86,10 +101,14 @@ type ExtensionBridge struct {
...
 	startErr error
 
-	mu       sync.Mutex // guards conn, connStop, pending
+	mu       sync.Mutex // guards conn, connStop, pending, pendingApprovals
 	conn     *websocket.Conn
 	connStop chan struct{}
 	pending  map[string]chan extInbound
+	// pendingApprovals maps an approval RequestID to the tool call awaiting a
+	// decision, so an approval_response frame can answer the right one. Normally
+	// holds at most one entry (the agent turn is paused while it waits).
```
edenreich:
```suggestion
```

internal/services/extension_bridge.go
```diff
@@ -34,12 +35,26 @@ type extInbound struct {
 	Token            string `json:"token,omitempty"`
 	ExtensionVersion string `json:"extension_version,omitempty"`
 	// browser_result fields (Content doubles as the user_message text)
```
edenreich:
```suggestion
```

PR #1064, opened 2026-08-14
fix(bridge): clear extension approval card on the answer, not the next event

edenreich:
## Problem

Tool-approval cards mirrored to the opentask extension panel appeared and then vanished milliseconds later, forcing the user to answer in the terminal.

The bridge's `chatPump` sent `approval_resolved` on **the first chat event after an `approval_request`**, assuming the agent turn blocks on the approval so nothing streams until the user answers. That assumption is false: `approval_request` takes a fast intercept path to the socket, while the tool-call chat events for the same turn are buffered through `RenderAGUI` and flush *after* it — each one tripping the "next event = answered" heuristic.

Confirmed via the extension's service-worker log:

```
approval_request         ← card shows
chat_event REASONING_END
chat_event TOOL_CALL_START/ARGS/END
approval_resolved        ← card cleared
```

## Fix

- Resolve on an explicit **`ToolApprovalResolvedEvent`** (new `domain` event, a `ChatEvent`), broadcast from `Coordinator.HandleToolApprovalResponse` — the single point where **both** terminal and panel answers converge. `chatPump` now emits `approval_resolved` only on that event; the speculative next-event heuristic is gone.
- Make `EventBridge.Publish` deliver control events reliably: streaming `ChatChunkEvent`s stay droppable (a lost delta is cosmetic), but a full buffer now **evicts the oldest chunk** to make room for a control frame instead of silently dropping it — so the approval (and the new resolved event) survive a long-response flood. Fully non-blocking, no cross-subscriber coupling.

## Tests

- `TestExtensionBridgeApprovalSurvivesTrailingEvents` — a trailing chat event must not resolve the card; only `ToolApprovalResolvedEvent` does.
- `TestPublish_DeliversControlEventWhenBufferFull` / `TestPublish_DropsChunksWhenBufferFull` — control events survive a saturated buffer; chunks stay droppable.

All pass with `-race`; `task precommit:run` clean.

Pairs with opentask#142 (the extension side).

### DIFF
--- a/internal/display/macos/event_bridge.go
+++ b/internal/display/macos/event_bridge.go
@@ -35,8 +35,8 @@ func NewEventBridge() *EventBridge {
 	}
 }
 
-// Publish broadcasts an event to all subscribers
-// Non-blocking: if a subscriber's channel is full, the event is dropped for that subscriber
+// Publish broadcasts an event to every subscriber. Delivery is non-blocking so
+// one slow subscriber can never stall the bus for the others.
 func (eb *EventBridge) Publish(event domain.ChatEvent) {
 	eb.subMutex.Lock()
 	eb.eventBuffer.Value = event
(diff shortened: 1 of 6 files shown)


internal/display/macos/event_bridge.go
```diff
@@ -45,16 +45,35 @@ func (eb *EventBridge) Publish(event domain.ChatEvent) {
...
+	select {
+	case s.ch <- event:
+		return
+	default:
+	}
+	if droppable {
+		return
+	}
+	// Buffer full. We hold s.mu, so no other producer adds and the consumer only
+	// drains - freeing a slot that stays free - so this send cannot block.
```
edenreich:
```suggestion
```

internal/display/macos/event_bridge.go
```diff
@@ -45,16 +45,35 @@ func (eb *EventBridge) Publish(event domain.ChatEvent) {
...
+	select {
+	case <-s.ch: // discard the oldest buffered event to make room
```
edenreich:
```suggestion
	case <-s.ch:
```

PR #1083, opened 2026-08-19
fix(computer-use): dock annotations, single-resample frames, saner frame retention

edenreich:
## Summary

Computer-use agents on macOS could not reliably click small UI like Dock icons. This PR fixes the whole grounding pipeline: frame retention, resampling, aspect-preserving coordinate mapping, OS-chrome annotation, a region zoom for native-resolution inspection, and annotation-parser robustness. Verified end to end: the agent zoomed into the Dock, identified Find My at exact coordinates, clicked it, and the app launched.

## Changes

- config/prompts.go: screen annotator prompt covers OS chrome (Dock icons, menu bar, desktop icons); GetLatestFrame description documents region zoom.
- internal/services/screenshot_server.go: single uniform resample from the native capture (was physical -> logical -> stretched 1024x768, losing Retina detail and skewing VLM grounding).
- config/computer_use.go: FitDims fits frames inside the target box with one scale factor (aspect preserved); default screenshot.buffer_size 5 -> 60 (frame files lived ~15s, shorter than the LLM round-trip that reads them).
- internal/agent/tools: MouseClick/MouseMove derive their API space from FitDims so capture and click mapping cannot disagree; GetLatestFrame accepts region {x,y,width,height} in frame space - the rectangle is re-captured at native resolution, annotated, and element coordinates are translated back into frame space.
- internal/display/resize.go: ResizeImage shared by the capture path and the zoom tool.
- internal/vlm: annotation parser salvages the summary and well-formed elements from malformed/truncated replies; default annotator max_tokens 1024 -> 4096 (dense annotations were truncated mid-JSON).

Follow-up: accessibility-tree grounding - #998.


internal/services/screenshot_server.go
```diff
@@ -223,22 +222,20 @@ func (s *ScreenshotServer) captureScreenshot() error {
...
-	targetW := s.cfg.ComputerUse.Screenshot.TargetWidth
-	targetH := s.cfg.ComputerUse.Screenshot.TargetHeight
-
-	if targetW > 0 && targetH > 0 {
-		img = resizeImage(img, targetW, targetH)
-		width = targetW
-		height = targetH
+	// Resize once from the native capture, uniformly on both axes: an
+	// intermediate physical->logical downsample loses Retina detail, and
+	// stretching to a fixed box skews VLM coordinate grounding.
```
edenreich:
```suggestion
```

